PR #11, opened 2020-04-07
Generic components

rmeyers4:
We should abstract out as much common logic as possible. Starting here with basic Form components and a Table.


frontend/src/Shelterly.js
```diff
@@ -27,9 +27,9 @@ function Shelterly() {
   }, [state.user, dispatch, cookies]);
 
   // Redirect to login page if no authenticated user object is present.
-  if (!state.user && !cookies.token) {
-    navigate('/login');
-  }
+  // if (!state.user && !cookies.token) {
+  //   navigate('/login');
+  // }
```
AlexMountain, line 32: Restore me!

frontend/src/components/Table.js
```diff
@@ -0,0 +1,72 @@
...
+          </tr>
+        ))}
+      </thead>
+      <tbody>
+        {rows.map((row, i) => {
+          prepareRow(row);
+          return (
+            <tr {...row.getRowProps()}>
+              {row.cells.map(cell => {
+                return <th {...cell.getCellProps()}>{cell.render('Cell')}</th>;
```
AlexMountain, line 62: Should this be a <td> instead of a <th>?

PR #15, opened 2020-04-17
Reactify hotline section

AlexMountain:
Convert Hotline Django pages into React


AlexMountain:
@rmeyers4 I think this is ready for a look!

PR #17, opened 2020-04-27
Reactify animals

rmeyers4:


AlexMountain:
@trevorskaggs I'm not seeing the "Add another" button when editing an animal?  Can you confirm on your end?

Added dynamic placeholder text, good idea!

trevorskaggs:
You are right! That issue disappeared after I repulled the newest changes.  Playing back through it again I did run into another issue though!
image
When I go through the Hotline - Owner Calling workflow, it seems to bonk when attempting to save the SR at the very end!

AlexMountain:
Good catch, fixed!

trevorskaggs:
Muy bueno! :shipit:

PR #20, opened 2020-04-27
Update Dockerfile

rmeyers4:


rmeyers4:
@trevorskaggs : ready for review. Should just be able to run `docker build .` from the shelterly dir. When you exec into the container you should be able to run the django dev server without having to source the virtualenv. The image is also in my Docker Hub repo with the `latest` tag if you would prefer to pull that down.

PR #46, opened 2020-07-13
Add notes models

trevorskaggs:
These are intended to track the series of communications and updates specific to an animal or owner.

The OwnerContactNote is to track when the owner has been contacted.  Examples of this would be immediately after an evac has been completed, when the animal needs to be sent to the vet, etc.

The AnimalNotes is to track specific issues that come up with an animal.  We will be tracking feeding, walking, pooping in a different Shelter Actions model.  This is specifically for noting anything specifically for medical, behavior, etc.


trevorskaggs:
Ready for review.

AlexMountain:
Can you add a description to this ticket explaining how/where you envision this being used in the UI?

trevorskaggs:
I will be making separate tickets for the actual workflows.  Just wanted to get the models built so they are there when the workflows pop up.

AlexMountain:
Well to be honest, I can't even evaluate these models without understanding their usage.

trevorskaggs:
Close as this will be handled in updated https://github.com/trevorskaggs/shelterly/issues/91

PR #49, opened 2020-07-16
Add basic python tests and get CI working

rmeyers4:
Add basic tests for DRF endpoints currently in use. Get tests to run with Gitlab CI.


rmeyers4:
Ready for review. Pushing any branch should now trigger python and javascript tests. I also changed how we are handling environmental variables vs. `secrets.json` to make it easier to connect to the test database.

PR #56, opened 2020-08-03
Add some initial Javascript tests

rmeyers4:
Add some basic JS unit tests.


rmeyers4:
@trevorskaggs : This is finally done... still working out the best way to mock API responses so we can test that the useEffect hooks populate our components w/ information as expected.

You can look at the code coverage info here. https://codecov.io/gh/trevorskaggs/shelterly/branches

trevorskaggs:
This looks awesome! Seems like a really good baseline for our to be building tests from.  The tool that you are using to monitor test coverage, codecov.io, is that a GH integration?  When this branch is merged, will all new branches be automatically published/review there?

Current View Branches View:
image

I also really like the way it allows you to sort for test coverage %:
image

As we know have this tool, are we thinking of setting medium term coverage % requirements for new master commits? @AlexMountain  you have input on that?

PR #65, opened 2020-08-10
Dev docker setup

### LINKED ISSUE #58: Add separate development image
The current prod image has some pain points for development. This should be fixed!

- [ ] Change entrypoint to start node dev server and django dev server by default

- [ ] Don't remove node_modules as part of docker build

- [ ] Add proxy to package.json for local development

rmeyers4:
Closes #58


rmeyers4:
I added documentation for development options here. The first two should work out of the box with the new docker image. The final one requires overrides that I will document separately.
https://github.com/trevorskaggs/shelterly/wiki/Development-Resources

trevorskaggs:
And this guy is ready for review?

rmeyers4:
@trevorskaggs : Ready for review. 

Checkout the branch locally and update your docker-compose to point to `trevorskaggs/shelterly:newdev`. 

Copy the docker-compose override template to where compose expects it.
`cp docker-compose.override.yaml.example docker-compose.override.yaml`

If you develop with a local git repo and want to mount it inside the container, leave the override as is. If you use VS Code to attach to the container using the git repo inside the container, update it to look like this:
```
version: '3'
services:
  shelterly:
    ports:
      - "80:80"
      - "8000:8000"
      - "3000:3000"
    entrypoint: tail -f /dev/null
    
```
Run the following:

```
docker pull trevorskaggs/shelterly:newdev
docker compose up -d
```
I added some bash aliases to make things easier. In the container, run `djangoserver` in one terminal and `npmserver` in another. You should be able to access the app at `localhost:3000`

If you want to test bundling the app into a js file and serving it out with the django dev server alone, run `deploy`. This will bundle the react app and move the staticfiles to the appropriate locations. Run `djangoserver` and you should be able to access the app at `localhost:8000`.

Issue #61, opened 2020-08-07
Add History tracking


AlexMountain:
We should add proper history tracking such that we always know the current users url history within our app.  Once in place we should start by updating and simplifying all of the back buttons. Looks possible leveraging history + the raviger package currently installed so we should start there.  If insufficient there are probably other packages out there that can do this.

- [ ] Replace back buttons
- [ ] Redirect to previous location when logging in if the user was logged out due to token expiration

PR #70, opened 2020-08-18
add back button to personform

### LINKED ISSUE #64: Add back button
Move the back button from card footers to the header.

rmeyers4:
Closes #64

### DIFF
--- a/frontend/src/people/PeopleForms.js
+++ b/frontend/src/people/PeopleForms.js
@@ -12,6 +12,9 @@ import * as Yup from 'yup';
 import { Switch } from 'formik-material-ui';
 import 'flatpickr/dist/themes/light.css';
 import { DateTimePicker, DropDown, TextInput } from '../components/Form';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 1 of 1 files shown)


rmeyers4:
@trevorskaggs : ready for review.

trevorskaggs:
Looking good!

image

It seems to take you back to the Hotline Landing Page when clicked.  If you are deeper into a workflow, is that still the intention? Do we want to figure out how to actually "go back a page" in the workflow? For instance when going Non-Owner and selecting "Add Owner", if on the Owner page you click the Back button, it took me back to the Hotline Landing.

AlexMountain:
@trevorskaggs Pretty sure it's just working the way it was before.  I think that is actually a pretty complex problem.  At first I thought it would just be a matter of implementing https://github.com/trevorskaggs/shelterly/issues/61, but I think it's even more than that.  Say the back button on the Owner page takes you to the Reporter, now what?  Is it a blank Reporter form?  or are we loading the previously created reporter data in?  If loading, now do we no longer want to be saving another new reporter, but instead just updating before moving onto Owner?  

Seems it will require some work probably best done in another ticket if we want it to do the above.

trevorskaggs:
Agreed.  This is good to go then!

PR #75, opened 2020-08-22
Lint js code

rmeyers4:
eslint


frontend/src/animals/AnimalForms.js
```diff
@@ -71,20 +79,20 @@ export const AnimalForm = ({id}) => {
 
   // Hook for initializing data.
   useEffect(() => {
-    let source = axios.CancelToken.source();
+    const source = axios.CancelToken.source();
```
trevorskaggs, line 82: Noticed these types of changes and that lead me to: https://ui.dev/var-let-const/#:~:text=Turns%20out%2C%20const%20is%20almost,it%20to%20a%20new%20value.&text=The%20take%20away%20above%20is,with%20const%20can't%20be.  I think these changes are okay, but we should note that the linter seems to prefer `const` in most cases.  It replaced `let` and `var` throughtout!

PR #76, opened 2020-08-25
Add status to SR Model

trevorskaggs:
- Add status column to SR model that has "Open", "Assigned", "Closed" as possible values
- Set default value to "open" so all newly created SRs are "open"
- Modify SR search page to allow for searching of all status and for "all" SRs


trevorskaggs:

PR #81, opened 2020-09-06
Evac assignments

trevorskaggs:
- Add Evacuation Assignment Model
- Build EvacTeamMember Endpoint
- Build EvacAssignment Endpoint
- Build Team Member Multiselect
- Send values (minus SR ids) to EA endpoint


trevorskaggs:
@AlexMountain Should be ready for review!

frontend/src/evac/AutoComplete.js
```diff
@@ -0,0 +1,171 @@
...
+class AutoComplete extends Component {
+  static propTypes = {
+    suggestions: PropTypes.instanceOf(Array)
+  };
+
+  static defaultProps = {
+    suggestions: []
+  };
+
+  constructor(props) {
```
AlexMountain, line 14: Want to take a stab at updating this example you used from the deprecated `constructor()` to a `useEffect()` hook?  see: https://reactjs.org/docs/hooks-effect.html#example-using-classes

frontend/src/evac/AutoComplete.js
```diff
@@ -0,0 +1,171 @@
...
+    super(props);
+
+      let source = axios.CancelToken.source();
+      const fetchEvacTeamMemberList = async () => {
+        // Fetch ServiceRequest data.
+        await axios.get('/evac/api/evacteammember/list', {
```
AlexMountain, line 20: Do we want to make this a generic file in our `Components` directory, with just a basic `fetchData()` function and I think all it would need passed in (through props?) is the url?

frontend/src/evac/EvacForms.js
```diff
@@ -158,9 +151,9 @@ export const TeamMemberForm = () => {
                   <Col xs={{size: 5, offset: 1}}>
                     <TextInput
                       type="text"
-                      label="Cell Phone*"
-                      name="cell_phone"
-                      id="cell_phone"
+                      label="Phone*"
+                      name="phone"
+                      id="phone"
```
AlexMountain, line 156: Thoughts on an additional button on this form to the effect of "Save and Add Another"?

trevorskaggs: Yea, that would be a good idea, and would be as simple as redirecting to the same page, right?

trevorskaggs:
This should be ready for review!
Requires:
`npm install`
`python manage.py migrate`

frontend/src/evac/EvacViews.js
```diff
@@ -1,21 +1,71 @@
...
   const [mapState, setMapState] = useState({});
   const [statusOptions, setStatusOptions] = useState({aco_required:false});
+  function getSelectedServiceRequests(){
+    var selectedIds = [];
+    for (var sr in mapState){
+      if (mapState[sr].hasOwnProperty('checked') & mapState[sr]['checked'] == true){
+      selectedIds.push(sr);
+      };
+    };
+    return selectedIds;
```
AlexMountain, line 30: I would probably simplify this function to be like so `return Object.keys(mapState).filter(key => mapState[key].checked === true)`

frontend/src/evac/EvacViews.js
```diff
@@ -1,21 +1,71 @@
...
+
+  // Team Selector Code
+  const [selected, setSelected] = useState([]);
+  const [teamData, setTeamData] = useState({options: [], isFetching: false});
+  function getSelectedTeamMembers(){
+    var selectedIds = [];
+    selected.forEach(function(item){
+      selectedIds.push(item.id);
+    })
+    return selectedIds;
```
AlexMountain, line 41: This function could also be reduced if you want to take a stab at it!

frontend/src/evac/EvacViews.js
```diff
@@ -126,115 +176,148 @@ export function Dispatch() {
...
-    <Container>
-      <Row>
-        <Col xs={12}>
-          <Map className="mx-auto d-block" bounds={data.bounds} onMoveEnd={onMove}>
-            <TileLayer
-              url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
-              attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
+    <Formik
+      initialValues={{
+        team_members: '',
```
AlexMountain, line 181: initial `team_members` should probably be an array?

frontend/src/evac/EvacViews.js
```diff
@@ -126,115 +176,148 @@ export function Dispatch() {
...
+      }}
+      onSubmit={(values, { setSubmitting }) => {
+        values.service_requests = getSelectedServiceRequests();
+        values.team_members = getSelectedTeamMembers();
```
AlexMountain, line 185: An alternative to simplifying your getselectedTeamMembers() code above could be to instead set the Formik field value for `team_members` using `setFieldValue()` when the input changes, rather than adding to the `selected` state.  Then team_members would just be already included in the submitted values.

AlexMountain (commented):
- [ ] Use two buttons when saving a team member.  (see animal form for forking logic on submit)
- [ ] Add evacteam and evacassignment to django admin.
- [ ] Thoughts on using black text for the team selector input?  The bright blue looks a little odd.
- [ ] Update team selector to use Formik field value instead of separate `selected` state tracker so that they are submitted in Formik's `values`.

PR #82, opened 2020-09-07
Address autocomplete using Google API.

AlexMountain:
Help suggest an address for users when filling out address fields.


AlexMountain:
- [ ] Capture lat/lon with google API return
- [ ] Disable original address fields
- [ ] Abstract into re-usable component

AlexMountain:
@trevorskaggs Should be ready for review!  See https://github.com/ErrorPro/react-google-autocomplete for instructions on setting the Google API key.

trevorskaggs:
Looking awesome!  Really love how this functions!  Couple notes:
- [ ] We need to figure out our strategy for the API Key component.  As we dont want to commit that to the repo, it currently requires a local change.  Looks like you can make/leverage as an environmental variable: 
```const loadScript = (url, callback) => {
  let script = document.createElement("script");
  script.type = "text/javascript";

  if (script.readyState) {
    script.onreadystatechange = function() {
      if (script.readyState === "loaded" || script.readyState === "complete") {
        script.onreadystatechange = null;
        callback();
      }
    };
  } else {
    script.onload = () => callback();
  }```
```  useEffect(() => {
    loadScript(
      `https://maps.googleapis.com/maps/api/js?key=${process.env.REACT_APP_GOOGLE_API_KEY}&libraries=places`,
      () => handleScriptLoad(setQuery, autoCompleteRef)
    );
  }, []);```
Source: https://medium.com/better-programming/the-best-practice-with-google-place-autocomplete-api-on-react-939211e8b4ce

- [ ] Also, it appears to clear out the other fields if you enter an address, then attempt to update to a second one. Seems to happen if you click or hit enter!

- [ ] When submitting during the Owner Workflow, at the end I get the following console error in Django:
```  File "/home/shelterly/shelterly/hotline/serializers.py", line 34, in to_internal_value
    data['latitude'] = float("%.6f" % data.get('latitude'))```

Note: the above does not seem to be a problem on Reporter or FR workflows!

trevorskaggs (approved):
Send it! :shipit:

PR #83, opened 2020-09-07
only log Django messages to console

### LINKED ISSUE #79: Fix logging issue in new Dev Container
I am getting some strange behavior with the new dev container:
image
When I try to use tab complete in the shell I get the above ^^

rmeyers4:
Closes #79 .

Due to an issue in downstream libraries, we are getting weird logging messages printed to ipython when trying to autocomplete. With this change we will only print Django errors in the console, rather than all printing all python errors.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -158,7 +158,7 @@ LOGGING = {
         },
     },
     'loggers': {
-        '': {
+        'django': {
             'handlers': ['console'],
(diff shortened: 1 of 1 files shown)


rmeyers4:
@trevorskaggs : Annoying bug, easy fix!

trevorskaggs (approved):
💥

PR #86, opened 2020-09-26
Evac dispatch map enhancements

AlexMountain:
- [x] Add a "Total Bar" at the top that tallies the total number of animals from all currently selected SRs
- [x] When a SR is selected and the map is moved so that SR is no longer in view, keep the SR in the list.  This allows users to not be confused on where these SRs have gone and for the above mentioned Total Bar to be more understandable in these scenarios.
-  [x] Figure out if we can stretch the map to fill the spaces on the left/right of the current frame, basically fill up the top section of the page.
- [x] Have SR Details links open in a new tab
- [x] Update look and feel of SR details page
- [x] Include status icons in the location popups
- [x] Add a zoom scale legend
- [x] 1 mile radius circle toggle
- [x] Update SR status when create an Evac Assignment
- [x] Prevent SR creation if open SR with same address already exists


AlexMountain:
@trevorskaggs I believe this one is in a good place for a review!  Punting on really polished icons for now, we can circle back with Font Awesome Pro in a later ticket.

evac/serializers.py
```diff
@@ -16,6 +16,9 @@ class Meta:
 
 class EvacAssignmentSerializer(serializers.ModelSerializer):
 
+    team_member_objects = EvacTeamMemberSerializer(source='team_members', required=False, read_only=True, many=True)
```
trevorskaggs, line 19: Do we want these to be `required=False`? I cant think of a case where we would want to allow that?

trevorskaggs (changes requested):
Nice! This is looking really good so far! A lot of really good quality of life improvements tucked in this bad boy.  Really only found a few small issues on the Dispatch Map page:

- [  ]Looks like we might have something wonky happening in the Dispatch map tooltip?
image
- [  ] Can we stretch the Team Selector object to fill that space up?
image
- [  ] Looks like we can have a race condition, if multiple people are deploying Service Requests.  Was able to generate the following: 
image
image
Think this is my bad for not limiting SR FK relationships, but we also dont want to as a SR can be assigned multiple times....  Maybe we just check and dont make the FK, if the SR is already assigned currently?  I can also ticket this separately if that makes more sense!
- [  ] If there are no unassigned SRs, we get this page:
image
- [  ] On the SR Details page, I am seeing the following in the console:
image
This doesnt seem to affect functionality, but could eventually!

AlexMountain:
@trevorskaggs Couple of things:

We can extend the team field to be the width of the page, but it bumps the DEPLOY button far to the right, kind of disrupting a compact flow.  I've made the change for the time being but we may want to circle on it again.

For some reason I can't seem to reply directly to your comment about the `required` flag, but those flags are specifically for the custom object fields, which we don't want to have to supply on PUT/PATCH/POST requests.  Whether or not the FK itself is required should still be controlled at the model level.

trevorskaggs:
> For some reason I can't seem to reply directly to your comment about the `required` flag, but those flags are specifically for the custom object fields, which we don't want to have to supply on PUT/PATCH/POST requests. Whether or not the FK itself is required should still be controlled at the model level.

Gotcha! That makes sense!

trevorskaggs (approved):
All requested changes retested, SEND IT!

PR #88, opened 2020-10-05
Use dynamic redirects for Back and Save actions

AlexMountain:
- [x] Replace form back buttons with the corner icon.
- [x] Make all back buttons dynamic using previous history location
- [x] Update object update form save calls to redirect to previous history location where relevant
- [x] Fix form Reset buttons to actually reset
- [x] Track and redirect to previous location when forced to login again.


AlexMountain:

trevorskaggs:
Everything is working as advertised! Forms reset, back buttons take you back a page in history!

The only problem I see so far is that in our multipage workflows, the basic back functionality is not going to be inutitive I think!

For example when doing Hotline > Owner Calling, you can enter the Owner information:
image
Then you would be starting to enter the animal information and the owner realized they gave you the wrong phone number (landline vs cell phone), so if you click back you are sent back to `/hotline/owner/new`, which is exactly where you were, but would seem more intuitive to send the user "back" to the Owner Edit view for the newly created owner! I know this is a whole next level, so I will extract this out to a different ticket!

In the mean time, can we hide the back buttons on pages 2 and beyond for our more complex workflows? Basically I think this is currently the Hotline/Intake flows! Button works as the user would expect on the first page though, so it can stay!

PR #89, opened 2020-10-12
Docker updates

trevorskaggs:
- Change Dockerfile to use a different username, to reduce confusion and fix mounting/overwritting issues.
- Standardize root settings to use os.path.join
- Add creation of default superuser to Dockerfile (admin/admin)


AlexMountain:
⚡ Good to go!

PR #95, opened 2020-10-20
Add ability to shelter and reunite animals

AlexMountain:
- [x] Add shelter room picker to animal intake add and animal update forms.
- [x] Add Reunification button to animal details that removes any shelter and sets that status to `REUNITED`
- [x] Add shelter info to animal details when it's sheltered.


AlexMountain:

animals/views.py
```diff
@@ -83,6 +83,9 @@ def perform_create(self, serializer):
 
     def perform_update(self, serializer):
         if serializer.is_valid():
+            # Set room to null if not present, or if status is changed to REUNITED
```
trevorskaggs, line 86: Are we checking the actual status here? I see that we are doing on the PATCH call from the form, but dont see the check here! Also can we do an update to status="Sheltered" if the room is added/updated?

trevorskaggs:
Just a few small changes to the `perform_create` method above ^^.  Also added labels to the Shelter dropdown.  Otherwise :shipit:

animals/views.py
AlexMountain: ah, I just need to clean up the comment.  I'm actually passing a null room when reuniting so it happens using the existing check.

AlexMountain: Updating status to SHELTERED seems important!  Added.

PR #111, opened 2020-11-03
Switched order of animal form fields

cdvieu:


trevorskaggs:
Ended up resolving this in https://github.com/trevorskaggs/shelterly/pull/123

Issue #104, opened 2020-10-30
Relabel "Evac" to "Dispatch"


trevorskaggs:
As we are learning that the Evac team will likely not be in Shelterly itself, we are going to move towards relabeling the current "Evac" module to "Dispatch"
- [ ] Update Sidebar
- [ ] Update router.js
- [ ] Update any relevant urls

mbokinala:
@trevorskaggs I can work on this can you assign it to me?

PR #116, opened 2020-11-09
Replace Dispatch In Menu

trevorskaggs:


trevorskaggs:
Ready for review.  Scaled back to just be cosmetics changes to not create conflicts! URLS will be fixed in #104

AlexMountain (approved):
@trevorskaggs Fixed up the tests and removed an unnecessary button while I was there.  If approved then good to merge!  🥫

PR #118, opened 2020-11-10
Add animal search

SpencerjCarroll:
Updated the animal search to use Card Groups, modeled after the service request search.


SpencerjCarroll:
wrong branch, we have two related to the animal search.

PR #121, opened 2020-11-12
Ea search page

cdvieu:
@AlexMountain @trevorskaggs - this is ready to be reviewed. 

As discussed, the only criteria missing is the ability to search EA's by Animal name, so we'll need to create another ticket to add that missing piece.


frontend/src/router.js
```diff
@@ -25,6 +25,7 @@ const routes = {
   "/animals/search": () => <AnimalSearch />,
   "/animals/animal/:id": ({id}) => <AnimalDetail id={id} />,
   "/evac": () => <Evac />,
+  "/evac/evacuationassignment/list": () => <EvacuationAssignmentList />,
```
AlexMountain, line 28: I liked the idea of referring to these as search instead of list. Let's update to `/evac/evacuationassignment/search` and `EvacuationAssignmentSearch`?

evac/views.py
```diff
@@ -29,3 +31,14 @@ def perform_create(self, serializer):
             action.send(self.request.user, verb='created evacuation assignment', target=evac_assignment)
             for service_request in service_requests:
                 action.send(self.request.user, verb='assigned service request', target=service_request)
+
+    def get_queryset(self):
+        queryset = EvacAssignment.objects.all()
+        status = self.request.query_params.get('status', '')
+        if status == "all":
+            return queryset
```
AlexMountain, line 39: Do we need this if check for "all"?  An unchanged queryset should return at the end as long as one of the other status conditions isn't met.

evac/views.py
```diff
@@ -29,3 +31,14 @@ def perform_create(self, serializer):
...
+        if status  == "open":
```
AlexMountain, line 40: extra space!

frontend/src/evac/EvacTables.js
```diff
@@ -52,3 +56,117 @@ export function EvacTeamTable() {
...
+                    </span>))}
+                </span>
+              </div>
+              <CardGroup>
+                <Card key={evacuation_assignment.id}>
+                  <Card.Body>
+                    <Card.Title>Service Request(s)</Card.Title>
+                    <ListGroup>
+                      <span>{evacuation_assignment.service_request_objects.map(service_request => (
+                          <ListGroup.Item key={service_request.id}>
```
AlexMountain, line 158: Let's get the SR address somewhere in here as that seems like an important identifier, maybe in a pipe section before the owner?

frontend/src/evac/EvacTables.js
```diff
@@ -52,3 +56,117 @@ export function EvacTeamTable() {
...
+        {data.evacuation_assignments.map(evacuation_assignment => (
+            <div key={evacuation_assignment.id} className="mt-3">
+              <div className="card-header">
+                Evacuation Assignment #{evacuation_assignment.id}<Link href={"/evac/summary/" + evacuation_assignment.id}> <FontAwesomeIcon icon={faClipboardList} inverse /></Link>
+                <div>Opened: <Moment format="LLL">{evacuation_assignment.start_time}</Moment></div>
+                <div>Closed: {evacuation_assignment.end_time ?
+                    <span>
+                      <Moment format="LLL">{evacuation_assignment.end_time}</Moment>
+                    </span> : "N/A"}
+                </div>
```
AlexMountain, line 145: I'm kind of curious if we can take better advantage of this space horizontally.  Maybe if we use a smaller datetime format, we can cram all 3 things here on 1 line between pipes? I'd give `format="lll"` a try.  Let me know what you think.

frontend/src/evac/EvacTables.js
```diff
@@ -52,3 +56,117 @@ export function EvacTeamTable() {
...
+                <span>Team Member(s): {evacuation_assignment.team_member_objects.map(member => (
+                    <span key={member.id}>
+                      {member.first_name} {member.last_name} &nbsp;
```
AlexMountain, line 148: Can we comma separate the team member names? see https://github.com/trevorskaggs/shelterly/blob/master/frontend/src/evac/EvacViews.js#L289 or https://stackoverflow.com/questions/23618744/rendering-comma-separated-list-of-links for an example

AlexMountain (commented):
Looking really good so far!  I suggested some minor changes but overall this is great.

frontend/src/evac/EvacTables.js
AlexMountain: I realize I'm probably the one that suggested only using animal count, but the trimmed down redesign feels like it gives us more to work with!  What do you think about breaking out the Animals section down to a 2nd line with a simple animal list of name/species, e.g. `Animals: Cosmos (Cat), Crono (Cat), etc`?  I think that should provide helpful identifiable information, @trevorskaggs can weigh in later too on his review.

frontend/src/evac/EvacTables.js
cdvieu: Changed the format to: `format="lll"` and consolidated the Open and Closed dates onto one line.

frontend/src/evac/EvacTables.js
cdvieu: Good call. I was meaning to do this and let it slip. Commas added.

frontend/src/router.js
cdvieu: Updated.

evac/views.py
cdvieu: Not needed. Updated.

AlexMountain (approved):
Looks like you need to merge in master and resolve a few conflicts.  Hit me up if you have any questions on doing that.  Once that's done this ticket is good to merge!  🔍

trevorskaggs:
This is AWESOME! Thanks @cdvieu! I'm realizing that perhaps sorting by date might be valuable, but in the short term, can we sort so that the newest as at the top of the results?

AlexMountain:
👍

PR #129, opened 2020-11-22
Update sr cards

cdvieu:


AlexMountain:
Sorry I'm a little late in digesting the requirements on this one.  The ticket looks great as per described in the issue!  However, I I think returning N rows per EA here won't scale very well once N gets > ~3, and I could easily see an SR site getting daily SIP assignments for a week+.  I spoke with Trevor and got some buy in.  Here are my proposed changes to the SR search card:

- [ ] Remove SR ID and SR datetime.  Replace datetime line with Address.
- [ ] Change Contact to be Information.  Keep Owner info here.  Only display Reporter info (in place of owner) if Owner is "Unknown".
- [ ] Instead of displaying all EAs, let's just add the info for a single EA if one is currently open for the SR. The SR serializer already should return an open EA ID if it exists, but it might need to be updated to also return the EA open date or just the entire EA object.  Add a close button that links to the EA resolution page (see EA search for example).  Let's move this into the Information block.
- [ ] Since Animals can also scale with N, I'd like to suggest moving the Animal list into it's own row, similar to how we're displaying them for SRs in the EA search.  I think in this case we will want to display some combination of name, species, and status per animal.

cdvieu:
@AlexMountain - addressed your PR comments. I could not access the Tombstone icon since it requires a Pro License, so I improvised and chose a free icon. Let me know if you think the layout needs any updates after reviewing. Thanks!

AlexMountain (approved):
@cdvieu These changes look great, approved!

trevorskaggs:
Yo! This is looking good so far! Only thing I noticed was the the fonticons are a different color than in other places!
image
Can we make those a white like the other ones for the time being?

cdvieu:
Updated blue to white. I was talking to Alex about that yesterday evening. I was just experimenting and looking for feedback. I thought maybe some color might help things standout a bit more, but we decided to save that type of work for a future issue.

trevorskaggs:
Cool, I love it! I think we can really sit down and think about how to make stuff more obvious, they for sure stood out!

PR #131, opened 2020-11-23
basic pagination bar, start showing/hiding based on page #

### LINKED ISSUE #124: Front End Pagination
We could use a front end pagination solution for a couple of areas of the site that can return large N numbers of objects:

- [ ] Action history tracking on object details pages
- [ ] Search pages

Hopefully we don't need anything too complex, but whatever we implement needs to look good!

rmeyers4:
Closes #124


frontend/src/hotline/HotlineTables.js
```diff
@@ -1,18 +1,22 @@
...
 import { Link } from 'raviger';
-import { Button, ButtonGroup, Card, CardGroup, Form, FormControl, InputGroup, ListGroup} from 'react-bootstrap';
+import { Button, ButtonGroup, Card, CardGroup, Form, FormControl, InputGroup, ListGroup, Pagination} from 'react-bootstrap';
 import Moment from 'react-moment';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import {
   faClipboardList
 } from '@fortawesome/free-solid-svg-icons';
 
+const ITEMS_PER_PAGE = 2
```
trevorskaggs, line 11: Do we want to maybe make this an Environmental variable? We thinking this might need to be different for each usage?

trevorskaggs (commented):
This works awesome! Only open questions that I have:
1. Can we make the ITEMS_PER_PAGE an environmental variable to be controlled? That or make it a global somewhere so that we have the same pagination value throughout an implementation!
2. Can we stylize the buttons at all? They work awesome, but the lime green clashes a bit with our current motif....

rmeyers4:
@trevorskaggs : I matched pagination buttons with the `btn-primary` colors we are using for the searchbar buttons. We can also use grey instead. 
I pulled out the # of items per page to a `constants.js` file.

trevorskaggs (approved):
SENDIT 💯

PR #132, opened 2020-11-24
Continued Map Enhancements

AlexMountain:
- [x] Need to make Map respect "Follow Up Date" constraint.
 - [x] Add ability to disregard "Follow Up Date" when needed. Add toggle next to ACO Required > "Pending Only" that will show/hide those that are currently before Follow Up Date.
 - [x] Lock Map to always be full viewable.


AlexMountain:

trevorskaggs:
Looking sharp!

- [ ] Map is staying fixed, but the stuff at the bottom seems to be able to scroll under.  Maybe we need to make the SR table its own container/scrollable?
image

AlexMountain:
@trevorskaggs I changed strategies from the sticky map to just making the SR list scrollable.  Definitely feels better.  Moved some other things around, let me know if you approve!

trevorskaggs (approved):
looks good!

PR #143, opened 2020-12-08
Dispatch Summary improvements.

AlexMountain:
- [x] Need to make `Owner Notified` not required in the case of non-owner animals. Might not even show this if the Owner is unknown.
- [x] Add "Closed" Timestamp
**EA Summary**
- [x] Hide "Close" button for Closed EAs, maybe this should just be an Edit button?
- [x] Fix initial state when editing a dispatch summary
- [x] When viewing a previously closed EA Summary, show the VisitNote for that instance and the "Owner Notified" Flag for each SR.
- [x] Need to add little Notebook icon to each animal on the DA Resolution Page, should open Animal Details page in a new tab.


AlexMountain:

trevorskaggs:
Looking good so far!  Pushed up a few Evacuation>Dispatch ui changes. One quick update:
-  [ ] Can we make an N/A or not show the `Notes` section if there are no notes for given DA/SR combo? Basically this only happens when you are looking at a freshly "Assigned" SR
image

AlexMountain:
@trevorskaggs Yes definitely, however, I'm in a bad spot migration wise so I'm going to merge this and remedy downstream.

PR #144, opened 2020-12-10
Add Owner M2M and update sections to support.

AlexMountain:
- [x] Convert Owner FK to a Owner M2M on SR
- [x] Convert Owner FK to a Owner M2M on Animal
- [x] Remove dummy "Unknown" Owner logic

**New workflows**
- [x] From a given SR, ability to add an additional Owner, this would add the SR and to each animal associated to the source SR
- [x] From a given Owner, ability to add an additional Owner, this would add to each animal associated to the source Owner
- [x] From a given Animal, ability to an additional Owner to just that animal, partial split custody cases.
- [x] Need to have the ability to orphan an animal that was assigned the incorrect Owner


AlexMountain:
@trevorskaggs Whew, ok, I think this one is ready for review!

trevorskaggs:
Yooooo.

- [ ] Animal Detail page seems to be bonking, guessing that the serializer needs some massaging in this case?
Additional detail: this was after the initial migration for existing data.  I was able to add a contact and the problem was fixed.  That being said, there is still the scenario where we "orphan" an animal where this would be a valid use case, or just a stray in general!
image

- Future looking problem is if someone adds the owner from multiple places you can end up with a spaghetti mess.  I was  currently able to add the same owner name to an SR and a specific Animal.  This ended up making 2 new owner objects as expected.  Not something to solve in this ticket, I need to learn how we want to resolve the de-dupes here.

Everything else is looking awesome, thanks!

AlexMountain:
@trevorskaggs Good catch when missing an owner.  An interesting point on dupes, at least if an SR has an owner, that owner should already be appearing on the animal, so if they can see an owner, hopefully they wouldn't dupe it?  I guess I could see them adding dupe individual owners on multiple animals.  I think if trained properly to add owner at SR level when appropriate, and the ability to remove owners from animals, at least gives them the power to clean it up?  Certainly something we can continue to think on.

trevorskaggs (approved):
Bug fix looks good, send it!

PR #146, opened 2020-12-14
Add OwnerContact Model, Hook into DA Resolution

trevorskaggs:
Based on feedback from LiveDemo/Intake discussion, it would be valuable to have an object that tracks what was communicated during each Owner Contact interaction.  This would be effectively a model that looks like this:
- [x] FK to Owner
- [x] Timestamp
- [x] Contact Notes
- [x] Optional FK to a specific Animal, this allows us to put a specific not when communicating about a given animal, if not populated, it will be assumed it applies to all animals owned by that Owner.

- [x] expose the Contact Notes field on the DA Closure page and generate an object for a given SR/Owner upon saving.


trevorskaggs:

hotline/serializers.py
```diff
@@ -104,6 +104,10 @@ class Meta:
 class ServiceRequestSerializer(SimpleServiceRequestSerializer):
     from people.serializers import SimplePersonSerializer
 
+    # Custom field to get Evacuation Assignments.
+    # def get_evacuation_assignments(self, obj):
+    #     return obj.evacuation_assignments.filter(service_requests=obj).values()
```
AlexMountain, line 109: Can you confirm why this was added?  In my testing I couldn't see any difference with or without it.  If we need it, please go ahead and uncomment, otherwise please remove.

people/models.py
```diff
@@ -22,6 +22,13 @@ def location_type(self):
     def __str__(self):
         return '{} {}'.format(self.first_name, self.last_name)
 
+class OwnerContact(models.Model):
+
+    owner = models.ForeignKey(Person, on_delete=models.CASCADE)
+    owner_contact_time = models.DateTimeField()
+    owner_contact_note = models.TextField(blank=False)
+    animal = models.ForeignKey('animals.Animal', blank=True, null=True, on_delete=models.CASCADE)
```
AlexMountain, line 30: Do we actually need this mapping to animal?  We're currently not using it anywhere and it seems kind of confusing to include and make available to changing in the update form.  Can the purpose of this mapping not be achieved with a proper note from the user mentioning which animals if any apply?

AlexMountain (commented):
@trevorskaggs Did some cleanup, almost there!  Just had a couple outstanding questions.

people/models.py
trevorskaggs: Originally, I wanted to have a per animal contact, but I think we will postpone that until we figure out that we need it.

hotline/serializers.py
trevorskaggs: Looks like this might be a merge issue, seems to have been added here: https://github.com/trevorskaggs/shelterly/pull/144/files#diff-4dd0495d31476abdadffd422687be444247200f0f690a6d6a0ada57c974067ddR90, will remove!

evac/views.py
```diff
@@ -85,7 +86,9 @@ def perform_update(self, serializer):
...
+                    VisitNote.objects.create(evac_assignment=evac_assignment, service_request=service_requests[0], date_completed=service_request['date_completed'], notes=service_request['notes'], forced_entry=service_request['forced_entry'])
                 else:
-                    VisitNote.objects.filter(evac_assignment=evac_assignment, service_request=service_requests[0]).update(date_completed=service_request['date_completed'], notes=service_request['notes'], owner_contacted=service_request['owner_contacted'], forced_entry=service_request['forced_entry'])
+                    VisitNote.objects.filter(evac_assignment=evac_assignment, service_request=service_requests[0]).update(date_completed=service_request['date_completed'], notes=service_request['notes'], forced_entry=service_request['forced_entry'])
+                for owner in service_requests[0].owner.all():
+                    OwnerContact.objects.create(owner=owner, owner_contact_note=service_request['owner_contact_note'], owner_contact_time=service_request['owner_contact_time'])
```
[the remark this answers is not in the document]
trevorskaggs: Valid point.  This would obviously work if there was only exactly 1 owner.  I will add a multiselect and then pass that value in.  This is also sloppy as I think it will be problematic for non-owner SRs...

trevorskaggs:
@AlexMountain Alright, I think this one is ready for review!

AlexMountain (approved):
Looks good!  I added the ability to populate these fields when updating a Dispatch Resolution. Please give it a cross test, then if good merge it!

PR #149, opened 2020-12-15
Owner search

cdvieu:
Opening a PR but there are few things that need to be addressed:
1.  Animals are associated to both Reporters and Owners but if an Animal has both, then there will be duplicate Animal entries. I need way to avoid duplicates on an OrderedDict
2. I need a way to figure out if a Reporter is a Reporter so I can add a condition to link to the Reporter Detail page, instead of the Owner Detail page.
3. Add tests


AlexMountain (approved):
@cdvieu Pushed up a round of minor changes, let me know if you have any questions.  Everything is lookin good so if you approve then good to merge!

cdvieu:
@AlexMountain  - nice work on polishing. Looks good to me!

PR #152, opened 2020-12-15
Action History Updates

AlexMountain:
- [x] Add "Change Reason" field to Owner Edit page
- [x] Track which Person fields were changed and what the previous/new values for those fields are.  I'm thinking we can do a JSON field here.
- [x] Add ability to track each time an Animals status changes and when that even occurred.


AlexMountain:

trevorskaggs (commented):
Found an interesting issue!  When removing an Owner from an Animal, I get the following in the "History" section: ```Trevor Skaggs changed this animal status to None 0 minutes ago.```

Everything else is looking cherry!

trevorskaggs (approved):
Looking good!

PR #154, opened 2020-12-18
Tie together EvacAssignments and Animals

AlexMountain:
- [x] Create relationship between EvacAssignment and the animals available for Dispatch, rather than all SR animals.
- [x] Filter animals on Dispatch pages to only be animals that were included in that Dispatch deployment.


AlexMountain:
@trevorskaggs I think this one is ready for review!  Requires a `migrate`

trevorskaggs (approved):
Really like this change.  Much better user experience and more intuitive!  Send it!

PR #175, opened 2020-12-30
Add functionality to create N copies of an animal.

AlexMountain:
- [x] Ability to mass copy an animal


AlexMountain:
@trevorskaggs I believe this one is ready for review!

trevorskaggs (commented):
This is very sexy!  Can we switch the place of `Age` and `No. of Copies` locations so the critical information is all up top for quick entry?  Also, I wonder if we should go with "Total Number" and start with a default of 1?  I was initially confused and thought I had found a bug, when I put 2 in and got 3 animals out!

This is going to save soooooo much time!

AlexMountain:
@trevorskaggs This was how I had it working at first but it felt weird when updating an animal and I thought making "copies" worked better for both instances.  I'll revert since it seems more intuitive, and just remove this capability from the update form.  I really like your idea on placement too.  Changes made, ready for another review!

trevorskaggs (approved):
Me and my 60 chickens approve of this ticket! 🐔 🇸🇳

PR #177, opened 2020-12-31
Add workflow stepper to Hotline SR creation

AlexMountain:
Add a stepper to the hotline workflow, it should achieve the following:

- [x] Provide guidance to the user through a multi-step/form workflow.
- [x] Include animal information to help when animal count is high
- [x] Allow the user to go backwards in the process and populate the forms with previously entered data.
- [x] Only create objects at the very end
- [x] Allow first responders to Add Owner


AlexMountain:
@trevorskaggs Let's get this bad boy reviewed!

trevorskaggs (commented):
Holy 🍙 🍙  this is awesome! Was able to walk through all of the scenario and everything seems to be working nicely!  The ability to go back and edit information throughout the process is just nails.

Couple of questions/suggestions:
- The colors seem to match okay, but get away from the dark theme a bit, are we able to control the style on that tool?
- When going back through the animals, if you have a few (I started to notice when I had 5+), it could be a big confusing to know where in the chain of animals you were. Not really critical, but a "nice to have" might be to indicated 3rd out of 5 or something to that effect.
- For the "Contacts" workflow, would it be possible to indicate "Owner Created" and/or "Reporter Created" instead of the count?  Might be nice to see that difference in the heads up.
- What do we think about graying out the "Next Step" button until the form has been touched? Or even better until the required fields are populated?  Not a huge issue, but I did try to click it before things where ready, but it obviously stopped me!

Nothing concrete in my findings, all the above are just ideas/suggestions.  Let's chat tomorrow and figure out if we want to make any of these before merging!

trevorskaggs (approved):
Per our conversation, we will spin off the stylistic changes into a later ticket. That being said, send this bad boy!

PR #188, opened 2021-01-07
Fix map performance at scale

AlexMountain:
Map performance needs some love at scale.


rmeyers4 (approved):
@AlexMountain : Works well, let's merge it up and I'll throw it onto AWS.

trevorskaggs (commented):
I was also able to pull down and was just in middle of making fake SRs to test scale, but if @rmeyers4 was already there, then :shipit:

PR #193, opened 2021-01-08
Display first responder agency

cdvieu:


trevorskaggs (approved):
Looking good, send it!

frontend/src/people/PeopleViews.js
```diff
@@ -70,7 +69,7 @@ export function PersonView({id}) {
             </Card.Title>
             <hr/>
             <ListGroup variant="flush" style={{marginTop:"-13px", marginBottom:"-13px"}}>
-              <ListGroup.Item><b>Name: </b>{data.first_name} {data.last_name}</ListGroup.Item>
+              <ListGroup.Item><b>Name: </b>{data.first_name} {data.last_name} {data.agency ? <span>({data.agency})</span>: ""}</ListGroup.Item>
```
AlexMountain, line 72: We actually have a line for `agency`  just a few lines down.  There's an argument for including agency in the name section for consistency but I think in this case where we have the space I like giving agency it's own line.  Maybe we should move it up so it's right below the name though?

Rest of the ticket looks golden!

cdvieu: Ah, I missed that. Good catch. Removed the Agency name from the Name field and moved the Agency field right below the Name field.

AlexMountain (approved):
Looks awesome!

PR #194, opened 2021-01-11
Enhance phone regex and clean phone field data.

AlexMountain:
- [x] Only store integers on the backend for phone numbers
- [x] Always display pretty phone numbers


AlexMountain:
@trevorskaggs I think this is ready for review!

trevorskaggs (approved):
This works amazing! I think we made the right choice on this strategy!

PR #196, opened 2021-01-11
Add found location

cdvieu:


cdvieu:
Running into a npm test failure but only after attempting to fix a merge conflict.

frontend/src/animals/AnimalForms.js
```diff
@@ -203,7 +221,18 @@ export const AnimalForm = (props) => {
             .nullable(),
           front_image: Yup.mixed(),
           side_image: Yup.mixed(),
-          extra_images: Yup.array()
+          extra_images: Yup.array(),
+          address: Yup.string(),
+          apartment: Yup.string()
+            .max(10, 'Must be 10 characters or less'),
```
AlexMountain, line 227: We decided to not use `apartment` for animal locations right?  Let's remove this from the Yup validation if so.

AlexMountain (approved):
One very minor nitpick, otherwise looks good!

frontend/src/animals/AnimalForms.js
cdvieu: Good catch. Removed.

trevorskaggs:
Couple notes from me:
- Can we expose the Found Location on the Edit Animal page also, might be nice if we need to update that!
- Let's add Found Location to what is being searched and to the Related Information pane on the right hand side.
image
- Seems like we might have lost the "Save" button in the Intake workflow? This might be resolved by @AlexMountain 's ticket though!
image
- On the Animals Details page, can we also add the Found Location is present to the `Location` pane? While we are there, lets change `Address` to `Current`.  So that its easy to see the difference when we then have `Found` right below it!
image

PR #201, opened 2021-01-13
Add stepper to intake workflow.

AlexMountain:
- [x] Add stepper to Intake workflow.
- [x] Add shelter FK to animals and add a way to pick a shelter to the Intake workflow.
- [x] Add workflow to room unroomed animals.


AlexMountain:

trevorskaggs:
Yo, I'm really liking the direction this ticket so far!  Couple of things that I think we should add in the near term:
- Can we add back the ability to specify the Room on the Animal Edit page?  Just might be a nice backup while we figure out exactly how we want the Drag/Drop functionality to work.
- Can we add the ability to specify Shelter during DAR? This would allow for a better interface/exchange between Evac/Intake/Shelter.  I'm thinking just a the same simple drop down that only shows up if you select "Evacuated" as the animal status.
- For the time being, can we have the name format for the Drag/Drop page be `Building Name - Room Name`? This will be a big bigger, but just makes it more clear for me when demoing this week!

Otherwise I think this is a really good start and should give us much to talk about during Intake.  I know that like we talked about we will need some kind of scale solution (i.e. group by building, dropdowns, etc.), but we can leave these decisions for later!  Lastly, I dont think we have to do this now, but I will get an idea during the demo of if we want a "Enter Shelter" layer for all Intake activities.  Basically allowing us to pass the Shelter PK through all Intake urls so we could be defaulting values if possible.

PR #202, opened 2021-01-15
Fix manifest and static files issue

trevorskaggs:
- [x] Fix issue with Manifest not loading correctly.
- [x] Fix url patterns for Static files so production will work!
- [x] Update gitignore to exclude directories we dont need!  Specifically the `static` folder in the root directly should just be where files are aggregated during `collectstatic`! All new static files should be added to `frontend/src/static`


trevorskaggs:
@AlexMountain I think this guy is ready for review!
I think you will need to do:
```
npm run dev
python manage.py collectstatic
```

AlexMountain:
Things seem good for me locally! @rmeyers4 Can you please test this one in a prod setting?

PR #203, opened 2021-01-17
modify datetime formats

SpencerjCarroll:


AlexMountain (approved):
Looks and works great!  My only question would be do we care to display seconds?  We weren't displaying them before, and I'm not sure we need that level of detail.  Obviously not hurting anything but might save a tiny it of space in places!  I will leave it up to you and Trevor to decide if we want to keep them or omit.

SpencerjCarroll:
Thanks for giving it a spin, oops yeah sorry about leaving the seconds in. Happy to remove them for now, dont really feel like that level of detail adds much.

PR #204, opened 2021-01-18
Update Django version in requirements.txt

SpencerjCarroll:


trevorskaggs (approved):
Seems like it didn't break anything obvious! :shipit:

PR #208, opened 2021-01-22
Add reporter name to animal search query.

AlexMountain:
- [x] Return animals in search if searching on reporter first_name or last_name


AlexMountain:

PR #224, opened 2021-01-31
Allow animal image uploads during Intake

AlexMountain:
- [x] Support animal image uploads during Intake.


AlexMountain:
@trevorskaggs This is ready for review!  Might be good to test it with the relatively larger images I sent you both locally and in a prod setting.

trevorskaggs (commented):
Ran into a weird thing:
- When there are multiple images the carousel doesn't seem to be working, and it doesnt seem to just be cosmos pictures. Single image is working alright still!
image
image

Otherwise tested all the images that you provided and did not have any problems on my local system:
image
image
Seemed to load up quickly!

Lastly, I did notice that I needed Debug=True for the images to display on Dev environment, but this the same issue that we have for Static files!

AlexMountain:
@trevorskaggs Hah, good catch, was a `<span>` issue!  Glad to hear uploads were fast.  Fixed and gonna merge.

PR #225, opened 2021-02-02
Restructure Shelterly

AlexMountain:
- [x] Create a single file for each page view.
- [x] Change owner -> owners on Animal/SR
- [x] Change evac -> dispatch on React side
- [x] Cleanup


trevorskaggs (commented):
Looking awesome.  Really appreciate you putting this ticket together.  Sooooo much easier to know where things are!  I went through the pages and only came up with a few small issues:
- [ ] Creating a new shelter, you get the following error when submitting:
image
- [ ] Shelter Dropdown on Animal Form seems to not be populating?
image
note: it does work on DAR page though!
image
- [ ] Not sure this is a recent issue, but I thought I was able to "unroom" animals in a previous version of the sheltering location.  Is that a change to not allow that?

PR #226, opened 2021-02-10
Add SR incomplete checkboxes to dispatch resolution form.

AlexMountain:
- [x] "Unable to Complete" - Allow for that SR to be skipped and put back into the pool with a status of Open.
- [x] "Not Completed Yet" - when selected, this would still save any changes that have been made but would not enforce that all animals be bucketed


AlexMountain:
@trevorskaggs Should be ready for review!  The only thought I have remaining on this ticket is whether or not we should be storing these values.  For example the case of "Unable to Complete" - it might be nice to report that information on the DA Summary page.

trevorskaggs (commented):
This ticket is looking good! Only issue that I found was when saving a "Not Completed Yet" SR, it Populates a Visit Note, or at least displays like there is a note!
Before: image
After Not Completed Yet: image
After Saving if I put in information into the Visit Notes Field:
image

In retrospect maybe we keep the note if they actually populate with a date and note value, but then if we dont, we dont want to make a blank entry?

AlexMountain:

trevorskaggs (approved):
Bueno!

PR #229, opened 2021-02-12
Assign an SR to an Existing DA

AlexMountain:
- [x] Assign an open SR to an open DA.
- [x] Transfer an assigned SR to a different open DA.
- [x] Add open DA map to Dispatch page


AlexMountain:
@trevorskaggs Ready for review!  Requires installing `@fortawesome/pro-duotone-svg-icons`

trevorskaggs (commented):
- Are we able to expose the same color circles right next the checkboxes for a given DA?
image
- Somewhat parrallel to what we are doing here, can we change this area to be:
`Active Dispatch Assignment - TIMESTAMP`
image
- Struggled to find the icon at first! Can we either add/move the icon up to the top with the Edit/Cancel functionality? I'm okay with it being in both places I think!
-Interesting result, if a DR has only 1 SR, the map is a bit confusing as it seems to select the current DR by default, which is likely covered by the Star!
image

Otherwise this is ANOTHER BIG WIN for the Dispatch teams!

AlexMountain:

AlexMountain:
@trevorskaggs Going to merge this so I can use some of the changes downstream without creating a mess.  Please feel free to retest and post any findings here to this ticket and I will be sure to address elsewhere.

PR #230, opened 2021-02-17
Shelter drag and drop improvements.

AlexMountain:
- [x] Highlight dragging object and droppable locations when active
- [x] Filter available rooms by building selection
- [x] Give rooms a fixed height and make scrollable
- [x] Handle long animal names

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -104,7 +104,7 @@ input[disabled] {
   padding-left: 0;
   margin-bottom: 0px;
   display: flex;
-  min-height: 25px;
+  min-height: 67px;
 }
 ::-webkit-scrollbar {display:none;}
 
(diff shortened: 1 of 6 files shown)


AlexMountain:

trevorskaggs (commented):
image
image
- Not sure if its related to this ticket, but I seem to be missing an animal?
- When we click the "Assign Animals" button from the Building/Room context, how hard would it be to start with that context on the Assign page? Basically auto-switching to the current Building?
- When there are no rooms associated to a given building, can we display a "No Rooms" type thing?
image

AlexMountain:
@trevorskaggs Ready for rereview!  I made an attempt to fix the missing animal bug but not sure since I couldn't reproduce.  If you're still seeing that ping me so we can track it down.

trevorskaggs (approved):
Full Send! 🚜 🦝

PR #235, opened 2021-02-21
Improve look and feel of main section pages.

AlexMountain:
- [x] Make the main sections for each app look consistent and more polished.


AlexMountain:
@trevorskaggs Ready for review, feel free to make or suggest any other tweaks!  Also, the intake page is starting to feel like it does less and less, hit me up if you want to chat about consolidating it!

trevorskaggs (commented):
image
Can we add the total number counts by species to this in some way?
image
Is there a way to either include DA# or Team Member Names on the tooltips here? Maybe we just put the Team # per our discussion earlier today once that is baked?  Might also be worth cards similar to Shelter section, but with the Open DAs?

AlexMountain:
@trevorskaggs Added the team members, good idea!  We can add the team name in addition once that's a thing.  On the shelter counts, I spent a few minutes trying to find a good place to display them but failed.  Rather than spending more time on it now, let's identify if we have any more reporting-type information we want for shelters and whether or not this is the page for it.

PR #241, opened 2021-02-25
Address Lookup Improvements

AlexMountain:
- [x] Add minimap to Address Lookups
- [x] Support cross street lookups
- [x] Support Lat Lon lookups
- [x] Draggable map marker to set precise location


AlexMountain:

trevorskaggs (approved):
This ticket is amazing and exactly what we needed!  Only tweak I would want is to not reset the zoom level after drag/dropping the pin!

AlexMountain:
@trevorskaggs I had played with better zoom handling but nothing was working so I moved on at the time.  Had an epiphany and took another stab due to your comment and I think I got it!  Gonna merge.

PR #243, opened 2021-02-26
Add Animal Index to Name

AlexMountain:
- [x] Add animal index when displaying an animal name throughout the site
- [x] Add animal list to reporter details
- [x] Update and polish all animal lists (person details, SR details, shelter details, room details)


AlexMountain:
@trevorskaggs I think this one is ready for review. Only page I deliberately skipped adding the animal ID to was the DA Search page because it still needs a larger overhaul.

trevorskaggs (approved):
Looking good! We might get future feedback on this, but for the time being, SEND IT! 🐟

PR #246, opened 2021-02-28
History pagination

SpencerjCarroll:


AlexMountain (commented):
@SpencerjCarroll Functionality seems to be working well but there's a spacing issue (larger gaps between history items)!  Also, the "key" you added also doesn't appear to be working as there's still a warning being thrown in the console (`Warning: Each child in a list should have a unique "key" prop`, see the unique key I used on the `Collapse`).  Do you even need to add a div here or can the `hidden` logic just be applied straight to the `Collapse` block?

AlexMountain:
@SpencerjCarroll Good to merge!  🌺

PR #248, opened 2021-03-02
Support dispatch team names

AlexMountain:
- [x] Add team name model migrations
- [x] Add dispatch team name through site when referencing a dispatch assignment
- [x] Support selecting a team in the dispatch map team member selector
- [x] Only provide team names from the previous day


AlexMountain:

trevorskaggs:
Getting an interesting result when adding a new member to an existing team, as we end up with multiple entries for "Team1".  I don't think this is actually wrong, just potentially confusing....
image
That being said, if you do the same process again, you end up with a true duplicate:
image

AlexMountain:
@trevorskaggs Do we effectively want the most recent "Team1" here?  I think we can do that fairly easily on the frontend.

AlexMountain:
@trevorskaggs Ready!

trevorskaggs (approved):
:shipit:

Issue #145, opened 2020-12-12
Filter Animals in Dispatch Assignment Resolution 


trevorskaggs:
Looks like we need to filter the list of animals displayed during this step to just those that are currently being serviced.  That would basically be any animals with either a status of Reported/UTL/SIP.  Other animals are currently returning that are listed as Evac/Shelter/etc. which can cause a accidental data entry issue!
![image](https://user-images.githubusercontent.com/1276123/101970887-f63ff880-3be1-11eb-9fd9-976b9fd1b738.png)
In the above example the team was just redeployed to look for the one UTL animal, but on resolution they are asked to triage all the original animals!

PR #257, opened 2021-03-09
Query linting

rmeyers4:


rmeyers4:
@AlexMountain @trevorskaggs : Ready for a review - I clicked through the app and don't think I broke anything. Still need to fix the Shelter module, but this should speed up most aspects of hotline and dispatch, and I'll tackle those after alex and I chat about ways we can maybe simplify our models.

evac/models.py
```diff
@@ -27,6 +27,7 @@ class EvacAssignment(models.Model):
 
     team = models.ForeignKey(DispatchTeam, on_delete=models.SET_NULL, blank=True, null=True)
     service_requests = models.ManyToManyField(ServiceRequest, related_name='evacuation_assignments')
+    # do we need this or can we get animals from SR?
```
AlexMountain, line 30: We added this mapping to make sure each DA had a static list of which animals were assigned when it was opened.  Unfortunately it can be the case where not all animals in an SR belong to a DA as animals can change dynamically on the SR. See https://github.com/trevorskaggs/shelterly/issues/145 for the original issue.  If axing this would help performance in any appreciable way, we can certainly try something else.  Let's talk about it tonight.

evac/views.py
```diff
@@ -54,7 +54,7 @@ class EvacAssignmentViewSet(viewsets.ModelViewSet):
     queryset = EvacAssignment.objects.all()
     search_fields = ['team__name', 'team__team_members__first_name', 'team__team_members__last_name', 'service_requests__owners__first_name', 'service_requests__owners__last_name', 'service_requests__address', 'service_requests__reporter__first_name', 'service_requests__reporter__last_name', 'animals__name']
     filter_backends = (filters.SearchFilter,)
-    permission_classes = [permissions.IsAuthenticated, ]
+    permission_classes = []
```
AlexMountain: Gonna add this back in!

hotline/views.py
```diff
@@ -12,11 +12,20 @@ class ServiceRequestViewSet(viewsets.ModelViewSet):
     queryset = ServiceRequest.objects.all()
     search_fields = ['address', 'city', 'animal__name', 'owners__first_name', 'owners__last_name', 'owners__address', 'owners__city', 'reporter__first_name', 'reporter__last_name']
     filter_backends = (filters.SearchFilter, filters.OrderingFilter)
-    permission_classes = [permissions.IsAuthenticated, ]
-    serializer_class = ServiceRequestSerializer
+    permission_classes = []
```
AlexMountain: and here!

urls.py
```diff
@@ -25,8 +40,9 @@
     url(r'login/', LoginView.as_view(), name='knox_login'),
     url(r'logout/', knox_views.LogoutView.as_view(), name='knox_logout'),
 ]
-urlpatterns += static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
-urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```
AlexMountain: Were these removed for a particular reason?  My media images broke without it.  Adding back but let me know if there's more to discuss here.

AlexMountain (approved):
Looking really good!  I found a few missing pieces of information that I added back in.  Please cross review my commit to make sure we don't need anything else to support my changes!  IAfter that then good to go!  🎭

PR #262, opened 2021-03-17
Move LoginForm to be on Home page.

AlexMountain:
- [x] Use Home url to be our login form page.


trevorskaggs (approved):
Groovy! 👍 Will circle back and double check the next time we push up to prod!

PR #264, opened 2021-03-19
Dispatch Assignment Search Improvements

AlexMountain:
- [x] Clean up and polish DA Search page results
- [x] Add minimap
- [x] Add pagination

### DIFF
--- a/evac/serializers.py
+++ b/evac/serializers.py
@@ -53,6 +53,14 @@ class EvacAssignmentSerializer(serializers.ModelSerializer):
     # action_history = serializers.SerializerMethodField()
     team_object = DispatchTeamSerializer(source='team', required=False, read_only=True)
     service_request_objects = DispatchServiceRequestSerializer(source='service_requests', required=False, read_only=True, many=True)
+    team_member_names = serializers.SerializerMethodField()
+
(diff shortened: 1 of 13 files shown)


AlexMountain:

trevorskaggs (commented):
Looking solid!

Might be an issue with `Unknown` parsing correctly? Seems to be present in the tooltip from the map and on the actual SR row on the right!
image
image

Are we able to expose the SR Details button? Also could we open up in a new tab if clicked on in the minimap? Not sure we "need it", but I think we offer that elsewhere?

- [x] When there is only 1 SR, we might need to zoom out as the tooltip actually gets clipped a bit!
Starting Zoom:
image
Zoomed out 4 Levels:
image
Zoomed out 1 level, but center offset slightly:
image

AlexMountain:
@trevorskaggs Changes made except for adding an SR details icon.  I'm hoping that by making the minimap SR icons clickable links we've achieved the desired functionality?  I've removed all other clipboard details links from our searches except for the object being searched on in an effort for cleanliness and consistency, but let me know if you feel we should make an exception here.  Otherwise ready for a rereview!

PR #272, opened 2021-03-26
Handle duplicate assigned SRs when deploying.

AlexMountain:
- [x] Reject Deployment if 1 or more SRs are already assigned.
- [x] Allow for user to reselect with duplicate offenders removed.
- [x] Allow for user to continue on with duplicate offenders removed.
- [x] Grab all SRs on initial load and use front end filtering for ACO and Pending options.


AlexMountain:

trevorskaggs:
Pop-up prompt came up in the right scenario, but I think maybe the SR details are missing from the modal?
image
Otherwise, looking awesome!

trevorskaggs:
Also ran into another small issue!  When selecting "Continue Without Duplicates", if all the SRs would be duplicates, you get the following error:
image
While this should not happen in most cases, if there is exactly 1 SR this could be a concern!

AlexMountain:
@trevorskaggs Not sure how I messed up the SR list.  Good catch on the empty submission. Ready for rereview!

PR #278, opened 2021-04-07
Animal medical notes field

AlexMountain:
- [x] Add medical_notes field to Animal model


AlexMountain:

trevorskaggs (approved):
Nice work.  I really like the tooltip detail on the search page, very subtle.  I will call that out on the next demo and see if we get feedback!

PR #315, opened 2021-04-29
Show assigned teams/team members and warn if dispatched.

AlexMountain:
- [x] Show already assigned icon in team member picker
- [x] Show modal warning when a user attempts to dispatch already assigned team members


AlexMountain:

PR #317, opened 2021-04-30
Da date range filter

cdvieu:


AlexMountain:
@cdvieu This is looking awesome!  Only issue I found is with the paginator not scaling down when the number of results are reduced, see example below with 1 result but 2 pages:

image

Want to take a stab at having it update accordingly?

AlexMountain:
Oh, I also just realized while looking at my own picture that we don't reset the filter when we clear the input.  We should probably do that too!

cdvieu:
@AlexMountain - regarding pagination - do you know if we take care of that scenario anywhere else in the app?

And, I'll look into the best way to clear the Date Range Filter!

AlexMountain:
I don't believe so but you should just need to recalculate and set the number of pages based off of the filtered array, using the same math as the initial set:  https://github.com/trevorskaggs/shelterly/blob/master/frontend/src/dispatch/DispatchSearch.js#L94

Maybe then just always set the current page back to 1?

PR #320, opened 2021-05-03
Use icons for "No" and False answers.

AlexMountain:


AlexMountain:
@trevorskaggs This is ready for review!  Static images will need to be added to S3 to show up.

trevorskaggs (approved):
SOLID

PR #322, opened 2021-05-14
Activate incident

trevorskaggs:


trevorskaggs:
@AlexMountain Hackish solution for the time being, but it works!

You will need to set an environmental variable inside your docker container (which prod deploys with), or just overwrite the `ORGANIZATION` value in settings.py

https://github.com/trevorskaggs/shelterly/blob/master/settings.py#L31

You can use the `test` organization.

AlexMountain (approved):
Code looks good!  Let's just remember to update the icon:  https://fontawesome.com/icons/siren-on?style=solid (note that it pulls from pro)

PR #324, opened 2021-05-18
use ref for search bar value

### LINKED ISSUE #323: Improve search bar responsiveness for long lists
When a character is entered in a search bar, on `onChange` event triggers a state update to `tempSearchTerm`, which causes the search bar and results to be re-rendered. When we have a relatively large number of results, this can lead to sluggish performance as several hundred re-renders occur every time a new character is entered. 

We can prevent the component from re-rendering after each character entry by making `tempSearchTerm` into a ref instead so component state will only update when the user clicks "search", triggering a re-render only when needed.

Ultimately we will probably need some sort of server-side pagination but this should work for the time being.

rmeyers4:
Closes #323 
Basically we will no longer be tracking `tempSearchTerm` in the react component state to prevent unnecessary re-renders.
We will only update the value of `searchTerm` when the user actually clicks Search.

There are probably better ways of accomplishing this by breaking the component into smaller chunks or having less results on the page by using server-side pagination, but I think this will be fine for now.

See this SO thread for more info:
https://stackoverflow.com/questions/50819260/react-input-onchange-lag

### DIFF
--- a/accounts/serializers.py
+++ b/accounts/serializers.py
@@ -7,4 +7,4 @@ class UserSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = User
-        fields = ('agency_id', 'cell_phone', 'first_name', 'id', 'last_name', 'username', 'email')
+        fields = ('agency_id', 'cell_phone', 'first_name', 'id', 'last_name', 'username', 'email', 'is_superuser')
--- a/animals/managers.py
+++ b/animals/managers.py
@@ -1,0 +1,10 @@
+from django.db import models
+from managers import ActionHistoryQueryset
+from ordered_model.models import OrderedModelQuerySet
+
+
+class AnimalQueryset(ActionHistoryQueryset, OrderedModelQuerySet):
+    def with_images(self):
+        return self.prefetch_related(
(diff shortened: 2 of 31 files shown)


rmeyers4:
@AlexMountain : Here's my proposed fix for the search bar issue we discussed tonight. If you approve we can apply to the other search bars as well.

AlexMountain:
@rmeyers4 I love the fix, it's noticeably snappier!  We should definitely merge this into all the search pages for 1.0.  My only question is, did you mean to gut half of the SR search result cards...?  (lines 213-249)

rmeyers4:
@AlexMountain : I did not! Remnants from experimentation. I'll add to the other search pages.

rmeyers4:
@AlexMountain @trevorskaggs : Finally got the conflict between the page and search term refs worked out, I don't think I broke anything else! Ready for review.

rmeyers4:
@AlexMountain : This is actually ready now, and I fixed the PersonSerializer that I broke for good measure.

AlexMountain:
@rmeyers4 Good to merge!  🔍

PR #348, opened 2021-09-01
Add full s3 support

trevorskaggs:


trevorskaggs:
@AlexMountain this ticket is not quite done yet! Need some react help to figure out how i can get into this logic: https://github.com/trevorskaggs/shelterly/blob/master/frontend/src/components/AnimalCards.js#L20.  However we approach it there are a few places to then copy/paste for the cards.  Basically I need that last section to be:

trevorskaggs:
Should be ready to review! Default you will be using S3, so you might need some env variable set to accommodate.  To use local static/media, you must have debug=True and use_s3=False!

AlexMountain (approved):
@trevorskaggs Seems to be working well locally, thanks!  If you have confirmed this works when deployed then ⚡

PR #363, opened 2022-01-26
Locking Django Version

trevorskaggs:
Seems like there is some incapability when Django was rolled to a new release.  Going to lock at 3.1 for now to fix test suite.


AlexMountain (approved):
Dig it thanks!

PR #376, opened 2022-02-23
Improvements to the Kennel Card pages.

AlexMountain:


AlexMountain:

trevorskaggs:
@AlexMountain This is 💣 .com! I know we are going to adjust margins in the future via the PDF method, so here is what a "print preview" looks like on my machine for the time being: 
image

Did it fill-out the bottom with action rows on your machine?

AlexMountain:
@trevorskaggs Interesting. I just did some more testing and it seems when I set my print margins to "minimum" it somehow manages to fill up the page with fewer rows. Changed back to default margins and I see the same you're seeing, so I'll go ahead and bump it up to 30.

I did notice one last thing though, our Shelter/Building/Room name string can be really long.  Is there any chance we could reduce this to just the Shelter name or the Room name?  If we need all 3 then we should probably rejigger things a little bit more to make room somewhere.

trevorskaggs:
Lets maybe just go with Building/Room in this instance?

AlexMountain:
Better than nothing!  I'll make the change.  Did you want to retest this one or is it basically good to merge?

trevorskaggs:
:shipit:

PR #377, opened 2022-02-24
Recheck the user authentication status when user refocuses the window

AlexMountain:


AlexMountain:
@trevorskaggs Should be ready for review.  I split out the Login page from the Home page in this ticket in order to have a simpler redirection use case when someone times out on the Home page so please do some general login/logout/refresh/session testing as a part of your review on this one.

trevorskaggs:
Looks awesome!

I test logging in/out and the time out at 1 minuter performed as expected.  They also requested a longer timeout this week, so I just bumped the default up to 8 hours to represent a "shift".  🚢 🇮🇹

trevorskaggs (approved):

PR #378, opened 2022-02-25
Handle duplicate owners

AlexMountain:
- [x] Handle duplicate owners when adding from the SR, Owner/Reporter, or Animal pages.
- [x] Handle duplicate owners when adding through the Intake stepper workflow.
- [x] Handle duplicate reporters when adding through Intake workflow by using existing Person if available.

### DIFF
--- a/animals/views.py
+++ b/animals/views.py
@@ -166,7 +166,7 @@ class AnimalViewSet(viewsets.ModelViewSet):
             images (List of AnimalImages)
         """        
         queryset = (
-            Animal.objects.with_images().with_history().exclude(status="CANCELED")
+            Animal.objects.with_images().with_history().exclude(status="CANCELED").distinct()
             .prefetch_related("owners")
             .select_related("reporter", "room", "request", "shelter")
         )        
--- a/frontend/src/animals/AnimalForm.js
+++ b/frontend/src/animals/AnimalForm.js
@@ -334,7 +334,7 @@ const AnimalForm = (props) => {
                 ]);
               }
               // Create Owner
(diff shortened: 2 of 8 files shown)


AlexMountain:

trevorskaggs:
Looking pretty good so far!

1. Was able to run into a little discrepancy on phone number check:
 
Original Owner:
image

When entering a new Hotline, with the phone number of `5315151214` I get the expected popup:
image

If I instead enter with dashes or parens (I found this the first time because I copy/pasted from the Owner Details page), then I dont get the popup:
image

I know we are doing some cool post formatting to that field as phone numbers are wonky AF. Wonder if we need to do that validation logic to the field before doing the dupe check?

2. Seems like we are checking for reporters also during Owner check?
Reporter Details: 
image
Dupe Check
image

Also fun fact: for a reporter the `Click here to view the owner` link will actually open a valid page as its just a person object and make it look like they are the owner!

image

AlexMountain:
@trevorskaggs 

1.  Good catch. I'll play around with it.
2. Yes that was intentional. Reporter objects are still person objects in our system and it made more sense to reuse the person if the information is identical.  It would technically make them an owner and reporter (but for different animals).  Are there any issues with this?

trevorskaggs:
2. Nah, I think the only problem I see is in that we say "Owned Animals" on the Reporter Details page.

AlexMountain:
@trevorskaggs Should be fixed.

trevorskaggs (approved):
Heck yea! :shipit:

PR #384, opened 2022-03-06
Better handling of active DAs when SRs/Animals are canceled.

AlexMountain:


AlexMountain:
@trevorskaggs It might be nice to have a quick discussion to make sure we're on the same page when it comes to showing/hiding objects but a quick summary is that I tried to hide all SRs with no animals and all DAs that have no requests (on the search, map, and summary pages).  This should be ready for review unless we want any further changes.

trevorskaggs:
Got something wonky going on in here.  

Creating a new DA I'm not able to view it:
image
image

Existing open DAs seem fine:
image

Closed/resolved DAs seem fine:
image
image

Search Page also seems fine:
image

I checked and I am able to view on master.

AlexMountain:
@trevorskaggs Seems like it was a bug with my change when multiple SRs are assigned and I was just testing with singles!  Pushed a distinct() and it seems to clear up the error if you'd like to retry.

trevorskaggs (approved):
Nice, looooooking good!

PR #385, opened 2022-03-07
Fix image removal/upload logic

trevorskaggs:


trevorskaggs:
Still have a weird issue when uploading a new Front/Side image when the other image already exists.  It causes a new file to be created with the same name `blob` and thus "updates" both images to the newly updated one.  This does not occur with the Extra Images field.

AlexMountain:
@trevorskaggs It's because we're only compressing front and size images atm.  It took some digging but I figured out how to set the filename when submitting a blob. Can you test it out for me in the cloud?

AlexMountain (approved):
Lookin good!

Issue #218, opened 2021-01-27
Develop Standard PDF Output of DA Evac Form


trevorskaggs:
To allow the Evac teams to get into the field with a paper version of the DAR, we need to design an easily printable and consistent format of output.   The form will look very similiar to the current DAR Form (Team Member Names, SR Details, Animals, Notes, Owner Contact, Etc.).  Main difference will be Status field will need to be Text in a line to allow for circling of correct status.

trevorskaggs:
@parksjr To generate a mockup based on the current DAR page: dispatch/resolution/<pk>

parksjr:
[DAR Form mock.pdf](https://github.com/trevorskaggs/shelterly/files/8191356/DAR.Form.mock.pdf)

Something like the attached mockup can be done using the jspdf library. Let me know if you want anything moved around or if I missed anything. @trevorskaggs

trevorskaggs:
This is coming along nicely! 

Can we add this information the SR header section:
![image](https://user-images.githubusercontent.com/1276123/156933457-ff1e4264-fa95-45d4-b91d-0e6362fe878a.png)
 (and grab the Owner Phone Number also after the name)

Also are we able to add the Owner Contact section please:
![image](https://user-images.githubusercontent.com/1276123/156933475-a499f20f-d4f9-4890-bc08-80a6cea1da3d.png)
For the Dropdown Owner Contacted, let just make a list as checkboxes!

parksjr:
@trevorskaggs thanks! I think the SR i created to test and mock around i didnt put an owner in, so I guess that pdf is what it would look like if there is no owner, if at all likely or possible. Anyway, i am attaching a new mock that includes the owner info. The page will likely always need a two page print out even with the minimum amount of info, so if you want me to try and cram things a little more I can, but not much room is there to begin with. Also a quick question: This pdf can be coupled to the ui elements and the data that is present on the UI, or do you envision having a separate endpoint to retrieve all of this data specifically for this pdf form on a click event? we can discuss next meet.

[DAR Form w owner.pdf](https://github.com/trevorskaggs/shelterly/files/8193855/DAR.Form.w.owner.pdf)

trevorskaggs:
[...]

PR #389, opened 2022-03-12
Issues/218 dar pdf download

trevorskaggs:


trevorskaggs:
@parksjr Can we move the "Print Button" next to the clipboard on this page?
image

parksjr:
Added the print icon/button for the offline resolution form to the Summary page (aslo fixed a bug with the animal grid on said pdf). I kept the button to download from the form as well, I figured why not have it in both places, but if you think it should only be on the Summary i can get rid of it on the online form, let me know either way.

trevorskaggs:
Client Updates:
- Make first page a summary page.  Need to include the following:
- SR: ID, Address, Animal Counts (2 Dogs, 1 Cat, etc.)
- Hard Page Break to cause the first SR to be on its own page

- For the Animal section would need to make a few updates/additions.
- Would we be able to maybe push this into a Table?  Something like this:
Animal ID | Animal Name | Species | SIP | UTL | Sheltered | Deceased
A#1           | Jeff                 | Cat        |  [ ]  |   [ ]  |      [X]       |         [ ] 
Primary Color:   Black, Secondary Color: White
Description: Has snowshoe markings.  Will typically be found under the back deck.
Behavior: Slightly human averse, can be coaxed with treats.
A#2           | Steve              | Dog       |  [X] |   [ ]  |     [ ]        |         [ ] 
Primary Color:   Tan, Secondary Color: Brown
Description: Looks like a German Shepard, very friendly
Behavior: N/A

- Priority Section:
- Can we make these checkboxes horizontal (maybe shrink font size)?
- Can we put a `(Current)` next to the level the SR currently is?

-Dates:
Can we put a format like "__ / __ / ____ (Month/Day/Year)" to help people on these fields?

Notes Fields
- They want to see at least 4 lines for to allow for longer notes.  That being said, the line spacing can be reduced to compact everything a bit more.

parksjr:
@trevorskaggs I just pushed up some changes that covers just about everything requested above. Only one thing remains, the Behavior Notes are not present in the animal object, so it is always N/A right now. I am not sure what the fix should be right now. Other than that everything should be as requested. 

closes #218

trevorskaggs:
Thanks for the update @parksjr ! I pulled down the most recent changes and straight away am getting: 
image
This is true for existing and newly created DAs!

trevorskaggs:
Also thought of another change that I didnt list before: can we have the file name be `DAR-###.pdf`?

parksjr:
@trevorskaggs sorry for the delay. couple updates here:

- was able to reproduce as you did, and fixed the bug
- pdf file name changed
- added the behavior_notes to the SimplAnimalSerializer as @AlexMountain suggested on standup
- rebased the branch to resolve conflicts
- easy peasy 😄 

if you see anything else let me know and Im on it

trevorskaggs:
@parksjr Thanks for the update duuuude!   Looking super clean.

Couple small things:
1. 
image
Looks like maybe we may need wrapping on Behavior (and maybe description field)?
2. I think I was not 100% clear on the Summary Page! I like the look, but I was thinking all the Summaries would be on the "Cover Page". Basically, can we take the details from pages 1 and 4 and drop them both on page 1?  See Attached DAR-007.pdf
[DAR-007.pdf](https://github.com/trevorskaggs/shelterly/files/8542812/DAR-007.pdf)
3. On the Summary Page, can we also add a team section at the very top? Basically just something like this:
Team Name: Team 2
Members: Trevor Skaggs (530) 232-2323, Mike Parks (215) 555-5523

Thanks dude!

trevorskaggs:
@parksjr THANKS DUDE! One last little tweak and then we will ship this guy! Can I get a line of space between the SRs on the summary page?
image

This is going to work awesome to get us out there with this functionality! Merge after the above change!!!!! :shipit:

parksjr:
And i just figured out how to deploy a team to multiple SRs. Thanks for testing this so well. I just pushed that last change up, I'll merge when all the checks pass.

PR #394, opened 2022-04-01
Disables and adds loading spinner on form buttons

parksjr:
- Created a new ButtonSpinner component, simply extends a normal button, includes a Spinner (react-bootstrap component)
- Replace <Button on forms with <ButtonSpinner

Could use a good eye. The AnimalForm and ServiceRequestForm were weird with the formik isSubmitting property so they're using their own state for that.


AlexMountain (approved):
Looks real good, merge on!

PR #397, opened 2022-04-11
Animal search filters

AlexMountain:


AlexMountain:
@trevorskaggs This should be completely ready for review now!

trevorskaggs:
Dude this works great! 2 Ideas/Request: 
1. Can we grey out the Apply button if nothing has changed on the form since the last time it was hit? So basically start with it Greyed out, add a filter, it becomes active.  Then if you apply that filter, it would grey out indicating it had already been executed?
2. Can we have a "No Results Found" box/indicator?

PR #402, opened 2022-04-22
adds mark.js library as a hook to highlight search terms on search pa…

parksjr:
Uses [mark.js](https://markjs.io/) to highlight search terms for Animal, Dispatch, SR and Owner search components.


trevorskaggs:
Thanks for getting this together! When you clear out the search field the highlighting sticks on the Animal Search!  This probably has to do with the search all being done on the front end now....
image
image

parksjr:
Ah, that one threw me off, for a few minutes, weird that only animals was doing that, but i figured it out. Good to go if you want to have another look

trevorskaggs (approved):
Full Send

PR #403, opened 2022-04-23
Support preplanned DAs.

AlexMountain:


AlexMountain:

AlexMountain:
@trevorskaggs I added the functionality to include teams in the existing DA team member selector so please make sure to thoroughly test that component as well, thanks!

trevorskaggs (commented):
Just killing it again.  How effective "preplanning a DA" is 💯 .  Only real comments I have are on style for this one:
image
1. On the Dispatch Landing page, we are using the same double ^ icon which is already being used for Priority indication, which was a bit confusing.  
2. Are we able to maybe go with a Dark Grey background for the Section Headers?  The blues are similar enough that I was thrown off initially.
3. On the Dispatch Landing page, lets change the the dispatch buttons to `Deploy Assignments` and `Preplan Assignments`

PR #406, opened 2022-05-10
Photo Document support for SRs/People.

AlexMountain:


AlexMountain:

trevorskaggs:
Coming along very nicely! Couple of comments:
image
1. Can we change to `Display Name` to `Name`
2. Can we change `Yes` to `Save`
3. When removing files, it appears to remove from the UI, but if I refresh the page the document is still there!
4. What does the lift look like to be able to edit a display name?
5. When clicking on an image, open up the image in a new tab.

After using this, its going to be super useful. Just going to capture some future work/improvements that we dont need to tackle right now, but if you can provide some insight on difficult/cost.
1. Add this functionality to the Animal page.  I think the Animal pictures are good to be stand alone, but having documents for an animal is going to come into scope as more Vet stuff is introduced.
2. What does support for PDFs end up looking like for this functionality? is the main hurdle how to display the document?

AlexMountain:
@trevorskaggs I bet #3 is a cloud vs local issue - I'll take a stab at it based on your other S3 image fixes. #4 is 30-60 minutes?

Adding to animals is a 1-2 hour job. Front end is re-usable and quick but I would need to make sure the backend plays nice with with the Animal carousel images.

Not sure what your asking about PDFs.  Do you want to be able to upload PDFs as documents?  or generating them from what's uploaded?

trevorskaggs:
Cool!
3. That is very likely!
4. Lets add this as I think it is worth the squeeze.

Adding Animals - Lets defer for the time being to a new ticket
PDF - Just the ability to upload a PDF in addition to JPGs.

AlexMountain:
@trevorskaggs Would you like to rereview this one?

AlexMountain:

trevorskaggs:
Just was testing Rename Functionality and found something a little weird:
1. Original Attachments:
image
2. Renaming one of the documents:
image
3. Results:
image

Those seem to be documents from a different SR....  If i refresh, I get back to the original documents:
image

trevorskaggs:
Additionally, those images from #3 above are actually on Owner #1 instead of SR #1 and the file has been renamed:
image

trevorskaggs:
After the above is resolved, lets send it!

PR #411, opened 2022-05-21
adds printable intake summary on click to owner details #404

### LINKED ISSUE #404: Intake Summary Printout
Objective of this would be to print a summary of the Owner Detail page.  The intention of this would be to have the ability to hand out the information directly to the owner after Intake.

image

Top would be the "Information" Section about the owner, then we would want to list out the animals attached to the owner.
For each animal, let's output the following: ID, Name, Species, Sex, Primary, Secondary Color, Breed/Description, Shelter Address.

parksjr:
Added the button to download a printable intake summary from the owner details screen, see attached pdf of an example of what that looks like (from my local environment). Let me know of any changes and i'll work it in.

[Intake-Summary-022 (32).pdf](https://github.com/trevorskaggs/shelterly/files/8748272/Intake-Summary-022.32.pdf)

Additionally, I've included the shelter_object to the ModestAnimalSerializer class, which  will have performance implications. The measure of those implications I don't know, but it's worth taking a gander at some point I'm sure instead of continuing to include fields like this down the road. At what point would it be a better remeasure to create a new serializer, or optional include?

Closes #404

### DIFF
--- a/animals/serializers.py
+++ b/animals/serializers.py
@@ -16,10 +16,11 @@ class ModestAnimalSerializer(SimpleAnimalSerializer):
     front_image = serializers.SerializerMethodField()
     side_image = serializers.SerializerMethodField()
     owner_names = serializers.StringRelatedField(source='owners', many=True, read_only=True)
+    shelter_object = SimpleShelterSerializer(source='shelter', required=False, read_only=True)
 
     class Meta:
         model = Animal
-        fields = ['id', 'name', 'species', 'aggressive', 'request', 'shelter', 'status', 'aco_required', 'color_notes', 'front_image', 'side_image', 'owner_names']
+        fields = ['id', 'name', 'species', 'aggressive', 'request', 'shelter_object', 'shelter', 'status', 'aco_required', 'color_notes', 'front_image', 'side_image', 'owner_names']
 
     def get_front_image(self, obj):
         try:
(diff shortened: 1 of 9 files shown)


trevorskaggs:
@AlexMountain Any opinion on the performance implications highlighted above?

trevorskaggs:
Coming along nicely! Now that I see it in person a few modifications (sorry):

image
1. Space between Service and Request in the Owner Details Section
2. Can we add a HR below `Owner Detail` to make that section have the same spacing/feel as the Animals section?
3. Let's add `Status` to the Animal Details section for each animal.
4. Can we Capitalize the species?
5. Last one: I realize that this is more of a Owner Summary now instead of a Intake Summary as this is going to be valuable for evacuated animals also when their owners arrive.  Can we rebrand that in the couple of locations: Tooltip on button, top of the report, filename?

parksjr:
latest commit addresses modifications, let me know how it looks.

trevorskaggs:
Thanks for the update, last little nitpicky things:
1. Can we add a space between Service and Request in the Owner Details section?
2. Standardize Case on Status Field (SHELTERED IN PLACE vs Sheltered In Place)
3. Can we fix the icon spacing at the top? Looks a little wide?
image

AlexMountain:
If we have performance problems I don't think it will be because of this. Probably easier to use as is for now and address as a part of a larger performance effort if we feel it's ever warranted.

parksjr:
@trevorskaggs , I fixed those last three things, I understood what you meant about a space between Service and Request this time 😆

trevorskaggs (approved):
Looking awesome, thanks for the changes!

PR #422, opened 2022-05-26
🐛  svg auto length bug fix

parksjr:
using "auto" as a height or width inline on an svg causes console error on chrome. fixed by using 100% instead.
image (1)


AlexMountain (approved):
Looks good thanks!

PR #423, opened 2022-05-28
Updates to Address Lookup component.

AlexMountain:
- [x] Add label to lat / lon map
- [x] Add error warning if address is not selected when searching
- [x] Add clear Button
- [x] Add tooltip to disable address fields


AlexMountain:

trevorskaggs:
Yoooo:
1. Looks like the Clear button doesn't clear out the Search box, intentional? If not, can we clear and move the window focus to the Search box?
2. When doing lat/lon, we don't get a search result to select? 
image

Error message and Map Not are 💯 .

trevorskaggs (approved):
:shipit:

PR #430, opened 2022-05-30
0.6.4 bug fixes.

AlexMountain:
- [x] fix sidebar display when window is shrunk.
- [x] fix DAR reported status error.
- [x] don't allow followup dates in the past


trevorskaggs:
@AlexMountain just need to extend the Followup Date Logic to the DAR page please!

PR #434, opened 2022-06-03
Adds lat/lon to dispatch assignment page and print

### LINKED ISSUE #415: Add Lat/Lon to DA Summary Document and DA Details Page
Need to add the Lat/Lon for each SR on the DA Summary Printout.

parksjr:
(closes #415)


trevorskaggs:
@parksjr Thanks for getting this one rocking!

Couple small issues I found:
1. Looks like our Status columns are a bit wonky?
image
2. On the First/Summary Page, can we put the Lat/Lon right below the address, similiar to how we have for the SR specific pages?
image

PR #438, opened 2022-06-15
Incident functionality

AlexMountain:


trevorskaggs:
As I was doing end-to-end testing, I'm just going to drop all my notes here and we can triage them into new tickets if need be!

Bugs:
1. Cant save an update to a Owner Contact Note:
image
2. Added an animal to an existing SR and the new animal does not get the "Contact" M2M populated:
image


Enhancements:
1. Can we order the animals by Animal ID in the Animal Block? Would apply to SR/Owner/Reporter Details Pages
image
4. Can we expose the DA number on the DA search page? Folks were searching by DA number and it wasnt as obvious since we have the search highlighting they were expecting that.
image
5. Let's require the "Change Reason" field when updating a Owner.
6. Errors on the Owner Contact Notes fields are pythonic
image
7. Can we add the "Spinner" to the "Save" button on the Photo Document modal?
8. When saving a new Evac Team Member, both buttons get spinners, not just the one you click!
9. When creating a new incident, can we making it so the default map location is the centroid of existing incident locations, but zoomed to maybe 7? Also can we allow clicking on the map to make the initial point? This would reduce the need for gathering the lat/lon if someone knew roughly where the incident was.  Once there is point on the map, drag/drop.
10. After draging/droping the Incident Pin, lets not rezoom the map.

Discussion Items:
1. You can get a weird scenario if you add an additional owner to a specific animal in a given SR.  You can then add an additional animal to that owner, but it doesnt end up actually going anywhere. You end up with an animal in "Reported" status, but not connected to a SR, so no one will ever actually work it. In images below Jeff is a dog in the original SR, and I added another owner, Frankie.  I then added an additional Cat directly to Frankie.
image
image
2. When a DA has multiple SRs near each other (or the map is zoomed out) you can get some conflict on the popups:

3. When a Evac Team Member is hidden, the teams they are in will still show up.
4. What do we thinking allowing "collapsing" on the DAR page? Just like a little accordion icon to show/hide the block and leave the header.

parksjr (commented):
@AlexMountain, I did a little code reviewing on this. The code is in a good readable state, even for me with not much Python experience. Here are a couple small things that stood out to me. Otherwise i give it all a big 👍 .
- /Incident/urls.py line 1, url is unused, maybe it will be used in the future?
- /Incident/tests.py is empty 😢

trevorskaggs:
Looking good, after a fix to the issue we discussed, :shipit:

PR #449, opened 2022-07-13
populates missing animal data on owner summary

parksjr:
(closes #446)

PR #453, opened 2022-07-14
Download SR summary as printable PDF

### LINKED ISSUE #448: Create SR Summary PDF

parksjr:
to be merged after #449 
closes #448

### DIFF
--- a/animals/serializers.py
+++ b/animals/serializers.py
@@ -21,7 +21,7 @@ class ModestAnimalSerializer(SimpleAnimalSerializer):
     class Meta:
         model = Animal
         fields = ['id', 'name', 'species', 'aggressive', 'request', 'shelter_object', 'shelter', 'status', 'aco_required', 'color_notes',
-        'front_image', 'side_image', 'owner_names', 'sex', 'size', 'age', 'pcolor', 'scolor']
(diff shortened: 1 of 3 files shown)


trevorskaggs:
Dope! Couple quick notes:
-  Can we just move this information down into the "Information" block: 
image
- Can we go a little horizontal with `Key Provided`, `Accessible`, and `Turn Around`?
- Lets add Breed/Description fields to the Animals.
- Can we get Primary/Secondary colors on the same line?
- If there are Behavior/Medical Notes, lets expose those, otherwise, dont show even the labels.
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

parksjr:
@trevorskaggs, here is what's in the latest update:
- moves id and priority to information block
- key, accessible and turn around formatted on one line
- breed / description is already included in the original commit, but the label will not be present if the breed / description (color_notes field) has no value
- primary/secondary colors formatted on one line
- behavior and medical notes fields are now added, and the lines will only show if a value is set (same as breed/description above)

parksjr:
I've fixed all merge conflicts and updated this branch on top of the latest from master, after the new incident functionality was merged in.

PR #461, opened 2022-07-25
Shelter Intake by DA.

AlexMountain:


AlexMountain:

trevorskaggs:
I really like the way this guy operates! One issue (maybe more of a question):

After sheltering an animal in a shelter (Shelter #1), if I go through the workflow for a different Shelter (Shelter #2), but same DA, I get a status of Sheltered, but with no Shelter selected.
image
Not really sure this is a problem, but do we want to show that the animal is already in Shelter #1?

AlexMountain:
@trevorskaggs This is a great catch, I think we definitely want to handle this better.  Should we display the other shelters here, or should we just be hiding those animals from the form altogether?

PR #465, opened 2022-08-02
Update animal species and color options

AlexMountain:


trevorskaggs:
Looks and works good!

One minor style change please:
image
Can we add a space between the species and color on the DAR and Intake from DA pages?

PR #469, opened 2022-08-05
Deceased Animal Report

AlexMountain:


trevorskaggs:
1. When cancelling out of Intake from Owner it takes me back to the Shelter page and then after a second, I get this error: 
image
2. When testing the Address Lookup fix on the Owner Hotline workflow (which works now!), I do see this error in the console: 
image
3. Zooming the Incident map works awesome! Can we allow the user to drop the pin on the map with a click if a pin doesnt already exist?
4. Can we change "Public" to "Shared" for Shelters?


Discussion Items:
1. So lets take this scenario: I make a Shelter in the test Incident and mark is a Public/Shared. Then make a new "Big Fire" incident, the Shelter will appear in the in the new incident.  From within the Big Fire incident, I can edit the shelter and turn off Public/Shared, which will drop it from the Big Fire incident! This was a bit confusing, so lets talk about how we want to address this workflow! I think either now showing the Public/Shared when not in the "parent" incident would be the easiest! We might need

AlexMountain:
@trevorskaggs Ready for rereview!  You'll need to update another FontAwesome package.

trevorskaggs (approved):
Full send! :shipit:

PR #488, opened 2022-09-08
Create global system error modal.

AlexMountain:


AlexMountain:
@trevorskaggs Should be ready for review! I tried to capture every API call except for the login/auth checks. To test I plugged in `0/0` to my API views in order to force failures.

trevorskaggs (approved):
I didnt get to each and every endpoint, but seems like the same approach throughout. :shipit:

Issue #459, opened 2022-07-20
Review NPM Vulnerabilities


trevorskaggs:
npm found 140 vulnerabilities (2 low, 40 moderate, 90 high, 8 critical)

PR #491, opened 2022-09-17
[#459] CVE cleanup

parksjr:
This was surprisingly easier than I expected, probably due to other projects we've worked on (no names) that had much more difficulty with the CVEs. Here are the steps I took to clean up the issues NPM audit reported, granted many of them are false positives, but  they are cleaned up none-the-less.

I ran `npm audit fix` as recommended by the `npm audit` command. That cleaned up some of the reported vulnerabilities, but was a small dent. Then I removed node_modules, reinstalled everything, started it up locally to browse around and test most of the familiar functionality, and everything still looked normal. 

Running `npm audit` again, it claimed that, "one of the vulnerabilities needs my attention." or something along those lines. I looked into it, it was about the `react-scripts` package, which is the main build tool for `create-react-app`. Our `react-scripts` package was a whole major version out of date, so I updated it, and that took care of all but 1 vulnerability report after running `npm audit`.

I proceeded to `rm -rf` the node_modules again to reinstall and start up to test functionality on my local machine, and everything still looks good. No weirdness, no errors, all good as far as I can tell. With the last remaining vulnerability, it was another "manual attention" issue, that had to do with the `nth-check` package which was many decedents deep under the `react-scripts` package, I manually updated that required version to the latest non-vulnerable version (^2.0.1) in the package-lock under the `css-select` dependency, and now we are at zero vulnerabilities. 

Finally, I imagine that somehow we can easily reintroduce the CVE's again, say if we delete the package-lock and regenerate it (which we all know sometimes happens in npm land) but we can cross that bridge when it's time to fix CVE's again, right? Right! Anyway, this needs a lot of QA to make sure there is nothing wrong. Thank you for attending my PR talk.


parksjr:
Getting a build issue with react-app-rewired static file names, I think we can work through this, but it has an impact on CI/CD, so I want to call a huddle with @trevorskaggs about some DevOps fun.

trevorskaggs:
@parksjr and I walked through this today and it was looking good! Not going to roll into 0.6.8, but lets merge it!

trevorskaggs (approved):

PR #504, opened 2022-11-11
Create base vet models.

AlexMountain:


AlexMountain:
@trevorskaggs This ticket is ready for review and could use some thorough testing!

trevorskaggs (commented):
Found one small bug and want to make sure I dont lose it! When working on a Treatment, if you save an Actual time, then go back into the Treatment and select "Not Administered" and then manually clear out the Actual Time, it will throw this error:

PR #523, opened 2023-02-03
Shelter intake summary

AlexMountain:


AlexMountain:

trevorskaggs:
Looking good so far!

1. Can we add the Owner/Reporter details/link in the Information Block in those workflows?
2. Looks like printing has an error right now! 
image
3. More of a Nice to Have: Can we have a Link from the Shelter to a List of Intakes from that Shelter? This could also possible be search page also, so you could go back to a Intake if you moved away!

trevorskaggs (commented):
Looking good so far!

1. Can we add the Owner/Reporter details/link in the Information Block in those workflows?
2. Looks like printing has an error right now! 
image
3. More of a Nice to Have: Can we have a Link from the Shelter to a List of Intakes from that Shelter? This could also possible be search page also, so you could go back to a Intake if you moved away!  Maybe just "Intakes Summaries" link in this block for now would be fairly simple? 
image

trevorskaggs:
3 above is OBE as it already exists!

AlexMountain:

trevorskaggs:
Looking super dope! Pushed up a an add for Admin on Intake Summary. Last small issue:
1. When intake using the Animal Copies functionality, you end up with an Intake Summary with only 1 copy of the animal.  The animal(s) are created on the back end and attached to the correct reporter/owner!
image
image

After that is fixed up, :shipit:

PR #525, opened 2023-02-05
Add Use Existing Owner lookup to Person form.

AlexMountain:


AlexMountain:

trevorskaggs:
image
Damn, that was fairly easy!

1. Can we update the Owner object if something is changed (phone number, address, etc), when the SR/Intake is completed/saved?
2. (Probably Follow-On Ticket) I think we can also now add animals to an existing SR if the SR is still open and we are reusing an owner.

trevorskaggs (approved):
:shipit:

PR #548, opened 2023-03-15
Email select users when SR is created.

AlexMountain:


AlexMountain:
@trevorskaggs Sorry for the delay. This is now ready for review!

trevorskaggs (approved):
Looks good! Tested all the bug Fixes and new UI elements, will need to test on UAT to confirm Email sends, but I see the code triggering locally and is the exact same behavior I get for new users emails locally.

Issue #550, opened 2023-03-22
Add "Intake Summary" PDF Print


trevorskaggs:
On the new "Intake Summary" pages, we are going to add a new print button in the Actions area that will package the following existing PDFs into a single PDF:
1. Owner Summary PDF (including liability release)
2. Animal Care Schedule PDFs for each animal.

If the intake has multiple owners (intake from DA), then we will do 1 and 2 above for each owner and include all in the single output.  Each Owner Summary needs to have its own distinct page numbering.

PR #567, opened 2023-04-19
adds print intake summary to pdf #550

parksjr:
The scenario I covered in here is

- navigate to shelter
- click intake either from owner or non-owner
- fill forms, submit
- print summary button at the top

Let me know if anything isn't as expected here or if I missed a step anywhere. I hide the print button if there are no animal or owner objects available, which I had when I did an intake from dispatch, but probably because I have a bunch of incomplete test dispatches I created or something or another.


trevorskaggs:
Thanks for charging hard on this!  Found a few little things!
1. Spacing on Page Title: 
image
2. Getting weird pagination issue on Liability Form from Owner Walk-In Flow:
image
3. Download button is not working on the Intake from Dispatch Assignment workflow!

trevorskaggs (approved):

PR #569, opened 2023-04-25
Issues/521 spinner icons

### LINKED ISSUE #521: Add Spinners To Non-Search Print Icons
With the exceptions of the Search pages, we need to add spinners to the other print icons as there is now a little longer delay due to the shrinking functionality being added.

Source: https://github.com/trevorskaggs/shelterly/pull/520

parksjr:
Everywhere there is a print icon, now has the spinner when clicked until the download starts. Added a new component to compartmentalize that functionality, along with some testing.

closes #521


trevorskaggs (approved):
This is soooo much better! #QualityOfLifeUpgrade

Issue #561, opened 2023-03-24
Print All on Animal Search is Triggering a Re-request of all animal images.


trevorskaggs:
Noticed this while testing as an "IPAD with Limited Bandwidth".  If you click "Print All" a new request is sent out for each of the animal images on the page, making for fairly bad performance with multiple images!

PR #574, opened 2023-05-29
[Issue #561] lazy loads image into data strings to prevent double loading

parksjr:
Implemented on animal search.

To test:

- open the network drawer for inspection on your favorite browser
- visit animal search, here you can filter and print all, print some, print one, however you like, do it all
- you should notices images load from network once, and any other times they will load as a data string which comes from from state

This does reduce bandwidth usage, however it does put more strain on the client, like more data in memory, weak clients will drag. Needs plenty of testing, especially to see how this loads with many images in the system.


parksjr:
This also needs some unit tests implemented, especially on the new code, but will cross that bridge later. this needs to get out there and manually tested out first.

Issue #580, opened 2023-06-13
Add Previous DA Visit Notes to DA Print


trevorskaggs:
To make sure evac has the information they need, add all previous DA Visit Notes and to the Summary section.  Add to the end of the SR Specific Section.

PR #596, opened 2023-06-24
#590 smaller font size and new fields on DAR print out

### LINKED ISSUE #590: Add Fields to DA
Make Font Smaller

Animal:
- Add Sex, Fixed, ACO Required, Aggressive, Confined, Injured, Last Seen, Age, Size
SR:
- Forced Entry

6/30 ADDITIONAL FIELDS
- Owner Alternate Phone Number
- Owner Comments/Alt Contact

-Key at Staging

parksjr:
AC from #590:
```
Make Font Smaller

Animal:

Add Sex, Fixed, ACO Required, Aggressive, Confined, Injured, Last Seen, Age, Size
SR:
Forced Entry
```

steps to QA:

- log in, select incident
- open search pane
- click Dispatch Assignments
- choose any single Dispatch Assignment from the listed results in the main pane
- print, and inspect the pdf
- rinse and repeat for as many DA's as possible

Closes #590

### DIFF
--- a/evac/serializers.py
+++ b/evac/serializers.py
@@ -54,8 +54,8 @@ class DispatchServiceRequestSerializer(SimpleServiceRequestSerializer):
 
     class Meta:
         model = ServiceRequest
-        fields = ['id', 'directions', 'latitude', 'longitude', 'full_address', 'followup_date', 'status', 'injured', 'priority',
-         'accessible', 'turn_around', 'animals', 'reported_animals', 'reported_evac', 'reported_sheltered_in_place', 'sheltered_in_place', 'unable_to_locate', 'aco_required',
+        fields = ['id', 'directions', 'latitude', 'longitude', 'full_address', 'followup_date', 'status', 'injured', 'priority', 'key_provided',
+        'accessible', 'turn_around', 'animals', 'reported_animals', 'reported_evac', 'reported_sheltered_in_place', 'sheltered_in_place', 'unable_to_locate', 'aco_required',
         'owner_contacts', 'owner_objects', 'owners', 'reporter_object', 'visit_notes']
 
 class AssignedRequestDispatchSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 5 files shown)


trevorskaggs:
Overall looking good! In the case of the new fields, some are left as blank, which is different then "Unknown" currently.  In those cases we are output blanks in the new fields, but N/A in older fields. Lets add the same N/A behavior to the new fields please!
image


Also if the lift isn't too big, can we get https://github.com/trevorskaggs/shelterly/issues/580 into this change set also?

parksjr:
> 

The default to "unknown" is back end functionality, should this be changed on the model, or should I put logic in the front end like:  `field === 'unknown' ? 'N/A' : field`

trevorskaggs:
The fields I'm looking at are Sex/Age/Size, which I dont think use the Unknown, so I think the check would be if the value is blank/null.

parksjr:
Latest update includes #590, #580, and #606.

For 580, I am not sure if that is exactly what is wanted, i see 1:1 SR->note for both owner contact and visit, but sounds like there should be an array of historical notes or something? Hit me up to discuss if needed.

trevorskaggs:
We are looking to grab all the "Visit Notes" from when an SR is assigned to a DA and completed. There should be a 1-n connection via the multiple potential SR-DA connections!

parksjr:
I think I got all the owner contacted notes in there now, but I still am not clear how the previous visit notes work.

parksjr:
I just got #603 added into this PR as well, plus an issue I discovered while looking at it, see comments in #603 for details. All that should be left is the visit notes which we just chatted about, I'll revisit tonight or tomorrow.

trevorskaggs (commented):
- [x] https://github.com/trevorskaggs/shelterly/issues/606
- [x] https://github.com/trevorskaggs/shelterly/issues/603
- [ ] https://github.com/trevorskaggs/shelterly/issues/590: Looks like this one is real close, just need to add: Owner Alternate Phone Number, Owner Comments/Alt Contact

trevorskaggs:
Upon further investigation, the fields are there:

trevorskaggs (approved):
:shipit:

PR #626, opened 2023-08-03
[#621] print on filtered SRs instead of all

### LINKED ISSUE #621: Date Filter does not apply to “Print All” for SR Search
If you apply a Date Filter, the search list appears to be filtered, but if you "Print All" you will get all of the SRs.  This behavior is not replicated on the "Download All" (GeoJSON) button.

parksjr:
Closes #621


trevorskaggs (approved):
:shipit:

PR #628, opened 2023-08-12
Add Organization model and data to owner liability forms.

AlexMountain:


AlexMountain:

I've added a dummy Organization in the migration as to avoid blow ups if one hasn't been created yet, but it will need to be changed first thing each upgrade/deployment if you want the forms to look right.

trevorskaggs (approved):
:shipit:

PR #629, opened 2023-08-12
Display modal if reset password token is expired.

AlexMountain:


AlexMountain:
@trevorskaggs This should be review for review! I kept things simple and added a modal requesting the user essentially start over. Let me know if you think anything further would be helpful or if we need to examine token expiration duration.

trevorskaggs (approved):
Bueno!

PR #643, opened 2023-09-20
Add export as CSV button to reports.

AlexMountain:


AlexMountain:
@trevorskaggs Let me know if you think the column headers returned in the CSV need to be improved, but otherwise should be ready for review!

trevorskaggs (approved):
Works Awesome! Only one small change:
image
Daily Report the order of the columns does not match the UI, I checked and all the other ones do!

PR #652, opened 2023-11-11
Expand Organization.

AlexMountain:


AlexMountain:
@parksjr Base models should all be added. I moved the home folder from accounts to incident. I left a TODO for the auth check since it seemed easier to validate my code once we're passing in the origination name.

AlexMountain:

trevorskaggs:
Was able to review above issues, new finds!
- When creating an initial Incident I get this error when loading the page:
```LatLngBounds.js:227 Uncaught TypeError: Cannot read properties of undefined (reading 'equals')
    at LatLngBounds.equals (LatLngBounds.js:227:1)
    at Map.shouldUpdateBounds (Map.js:281:1)
    at Map.updateLeafletElement (Map.js:135:1)
    at Map.componentDidUpdate (Map.js:254:1)
    at commitLifeCycles (react-dom.development.js:19835:1)
    at commitLayoutEffects (react-dom.development.js:22803:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237:1)
    at invokeGuardedCallback (react-dom.development.js:292:1)
    at commitRootImpl (react-dom.development.js:22541:1)
```
- Error thrown when going to Shelter Details in Training incident that did not create the Shelter. Room/Building seems to function correctly!
- Need to filter Owners/Reporter Search to only those that have animals in the current Incident.
- As a non-superuser and non-staff user I can access the User Management panel.
image

AlexMountain:
@trevorskaggs Good finds. Should be fixed. I went ahead and hid the User Management pages from people without permissions but I do want to note we have protections on the backend for POST/PUT/DELETE methods. I also added a ditty to add all superusers to an Organization on creation.

trevorskaggs:
Nice fixes! Only one more minor tweak for now, then lets merge this bad boy!
- Seems you fixed the Owners not showing up in Incidents without animals, but that is not true if i remove the "Owners" filter, they show back up, with no animals! I think this has to do with the defaulting to Reporter == Non-Owner. If you filter for either Reporter or "All" on the Owners search you will see all the Owners/Reports from the other incident!

PR #655, opened 2024-01-04
Complete the Vet Stepper Workflow

AlexMountain:


AlexMountain:
@trevorskaggs This one is ready for review! I would recommend deleting all vet objects, migrating down, and then back up if you were on a previous branch of this ticket.

trevorskaggs (commented):
Nice work on this dude!  Here are the things I found:
- Age Confirmation allows you to Confirm with blank values in the fields
- Microchip starts as in the "Present" state
- Can we bump up to the upper most error when trying to submit an exam with an issue?
- Looks like minor Alignment issue with the top line of the Treatments/Diagnostics/Procedures cards?
image
image
- When making a second VR, the Diagnosis from the first VR seems to pre-populate during the initial Exam/Workflow. (this is probably going to go away when we resolve where/when diagnosis comes up, but might be an easy fix for now!
image
- Diagnosis will not let you save and finish without at least 1 entry during Exam/Workflow.


Future Issues:
- Add ability to cancel Treatments/Diagnostics/Procedures in the cases of duplicates.

AlexMountain:
@trevorskaggs 
- [x] Age Confirmation allows you to Confirm with blank values in the fields - fixed
- [ ] Microchip starts as in the "Present" state - this was actually the requested default as per the defaults sheet. Let's ask about it.
- [x] Can we bump up to the upper most error when trying to submit an exam with an issue? - we were actually already doing this! ...Except for the toggle switches. Fixed.
- [ ] Looks like minor Alignment issue with the top line of the Treatments/Diagnostics/Procedures cards? I couldn't reproduce this one but I attempted to clean up some wonky css that will hopefully help.
- [x] When making a second VR, the Diagnosis from the first VR seems to pre-populate during the initial Exam/Workflow. - fixed
- [x] Diagnosis will not let you save and finish without at least 1 entry during Exam/Workflow. - fixed
- [x] Add ability to cancel Treatments/Diagnostics/Procedures in the cases of duplicates. - good suggestion. I had meant to come back to this later since we didn't have an already supported area to copy but I forgot. Fixed.
- [x] Added some bonus icons.

Want to double check this stuff before we merge?

trevorskaggs (commented):
Small things, then lets merge!
- Error when not selecting a Assignee is a bit wonky:
image
- Did we change the exam fields to be now be alphabetical? Was this requested/intentional?

AlexMountain:
@trevorskaggs I did change it to alphabetical intentionally. I noticed when updating my values that the order was constantly rearranging which was super annoying. Without implementing a full blown "priority" ranking system (that I can't imagine working well through Admin management), I considered sorting by ID or Alphabetical. ID seemed just as arbitrary and inflexible with new additions as Alphabetical but I felt Alphabetical had a clean and consistent look. Let's ask about it on our requirements call.

trevorskaggs (approved):

PR #656, opened 2024-01-12
Add Total table footers to respective reports.

AlexMountain:


trevorskaggs (approved):
Ticket works awesome! Only issue I found is that when exporting a report with a Total row, you get an extra column called `last`, with True/False values.  Lets get that out of the export, then :shipit:

AlexMountain:
@trevorskaggs Good catch. I also fixed up the user admin stuff.

PR #660, opened 2024-02-10
Issues/556 disable data dependant buttons

### LINKED ISSUE #556: Disable FE Action Buttons (Print/GeoJson/Etc.) Before Page Is Loaded
If you are able to click a FE (ones that are using the data to generate something and not load a new page) Action Button before the page loads, you will get an error.

parksjr:
Applied the fix to the following areas:

- animal details
- owner details
- service requests
- dispatch summaries

Here is a visual of what it looks like on the animal details, note the " - " in place of the id number, i put this there instead of being empty to soften the jump that occurs once it loads.

Screen Shot 2024-02-10 at 1 04 31 PM

Closes #556

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -21,3 +21,5 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+.npmrc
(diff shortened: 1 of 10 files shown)


trevorskaggs (commented):
You did exactly what I said! That being said, lets actually just blank all the buttons as i think will make more sense for the user even though "technically" they would work! Its a bit weird seeing a few buttons and I can predict the "why are these different" questions!

Also, can we slap this logic on the VR page as that is a new one with a few actions at the top!

parksjr:
LoadingLinkified all the things, including VR details page

trevorskaggs (approved):
:shipit:

PR #666, opened 2024-02-21
Vet landing and search pages

AlexMountain:


trevorskaggs:
In the current version of this ticket if you have not created an MR (via the creation of a VR), the animal will not allow editing.  You get the following error: `{"medical_record":["Incorrect type. Expected pk value, received str."]}`.  Realize this is not final, but kind of a tricky scenario and wanted to make sure we document it!

AlexMountain:
@trevorskaggs Sorry for the delay. Let's get the review goin on this one!

trevorskaggs:
Alright, found a few issues so far:
1. Can't save an animal until an MR is created as I get this error: `{"medical_record":["Incorrect type. Expected pk value, received str."]}`
2. Cant create a new VR as I get this error:

trevorskaggs:
- [x] https://github.com/trevorskaggs/shelterly/issues/677
- [ ] https://github.com/trevorskaggs/shelterly/issues/674
- [ ] https://github.com/trevorskaggs/shelterly/issues/673
- [ ] https://github.com/trevorskaggs/shelterly/issues/672
- [ ] https://github.com/trevorskaggs/shelterly/issues/668

AlexMountain:
@trevorskaggs I kinda had to rejigger the model to be properly directional to cleanup the animals with no MRs saving. So you will need to migrate and all existing MRs are gonna be blown away. VR form error was missing a presenting complaint - I've added the proper form support to make it required. Should be ready again.

trevorskaggs (approved):
Looking good!
- For Microchip search it works, but do we want to expose and highlight to show its working?
- For Treatment Notes, do we want to expose this on the Treatment Card?
- Respiratory Rate was requested to be a freeform text, but still seems to be numbers only: 
image
- Add Create VR Button to MR: looks like this one slipped through the cracks?
- Add `Temperature, Pulse and Respiration` to Treatments: I think this is all a proper known, but I think is configurable, so no problem there!

PR #667, opened 2024-02-22
fits radius bounds when adding point to animal advanced search map #462

### LINKED ISSUE #462: Zoom out to extent of Circle when selecting a Range Radius on Animal Advanced Search Map

parksjr:
Navigate to Search > Animals, click the Advanced button to expand search filters, click on map to add point and circle and watch it fit the bounds. change radius to and fro, change position, etc...

Screen Shot 2024-02-21 at 10 12 59 PM

Closes #462


trevorskaggs (commented):
Couple weird things im seeing in this branch...
1. I can seem to click and drag the map once a pin is pinned to the map, makes it super hard to move around!
2. Minor regression in that the map previously would zoom into near the incident (or maybe an aggregation of the valid SR/Animal points), but now seems to drop us zoomed all the way out on the whole world!

parksjr:
> 1. I can seem to click and drag the map once a pin is pinned to the map, makes it super hard to move around!

I removed the setting on the map that prevented dragging. Seems like that was intentional at some point, but removing it and testing it works fine, and sounds like that's the user experience we want, so let me know how that looks for you.

> 2. Minor regression in that the map previously would zoom into near the incident (or maybe an aggregation of the valid SR/Animal points), but now seems to drop us zoomed all the way out on the whole world!

The map is supposed to fit to a bounding box around all the defined found locations of the animals (not incident, or SR location). It looks like if only 1 animal has a lat/lng then it works as you describe, but if more than one animal has lat/lngs then it gives you the whole world. I messed around with it but couldn't figure out a solution. 

I think we should wrap this one up, and put something in the backlog to figure out the map zoom on load issue.

parksjr:
I have another attempt at a fix for this ready for review. It includes incident coordinates for the advanced search map view consideration. Let me know how it goes, @trevorskaggs

PR #695, opened 2024-03-19
adds pdf upload for Photo Documents for SR/Owner, and extra images for animal

### LINKED ISSUE #670: Allow PDF Uploads to Photo Upload
Currently it appears that we are only allowing images to be upload via the Photo/Document upload.  Seems like this is okay for Front/Side photos of Animals, but would be good for SR/Animal/Owner objects to allow the upload of multi-page PDFS.

Note: This may complicate our thumbnail/preview functionality.  Additional testing should also be done to insure we don't break Animal Care Card generation and the Animal picture carousal.

parksjr:
### Functional overview

For the Photo Documents component, the ImageUploading component was swapped out for FIlesUploading component of the same dev group. Front and side images remain the same image only component, while extra images is swapped for the same FilesUploading component used in Photo Documents.

### UX stuff

On the animal details page, using a function to distinguish images/documents that may exist as "extra_images" and filtering them respectively, and a new ListGroup section on the left pain will house any documents, while images remain in the carousel.

On the Photo Documents component (SR or Owner), the thumbnails will show a typical image thumbnail, or PDF icon if its not an image.

### More notes

Image compression still works the same for an image being uploaded, no compression is done on a file that is not an image.

Closes #670


trevorskaggs (commented):
Looking super doooooppee! Have 2 small fixups then, lets merge this bad boi:
1. For actual images we have lost the image preview 
image
image
2. When editing the PDF name, the preview image seems to be broken, I'm guessing we just need to sub in the PDF placeholder image!
image

parksjr:
@trevorskaggs, Sorry for the two-week delay! Both bugs mentioned should be fixed now. Bolstered up some tests around it, too. I'll keep my eye out for any replies in case something is wonky.

trevorskaggs (approved):
:shipit:

PR #718, opened 2024-04-17
Adds hide feature to incidents

parksjr:
- adds hide button on incident edit form
- adds hide endpoint (still WIP)

**Couple of issues I am having**

1. the endpoint isn't working for me, it's just going to 403 forbidden.
2. looked into running the tests on the python end, after I added my new migration for the new hide column, the test suite complains about that column not existing. Googling led me almost nowhere, I attempted to reset my migrations, it got weird, so I aborted that.

I'll be reaching out for help on the above next chance I get ✌️


parksjr:
One of the reasons I wanted to push this up was to see if CI hit the same wall I did on the column not existing, and it did, so there's that

parksjr:
@AlexMountain @trevorskaggs 

Thanks for the help earlier, Alex, the trailing slash was exactly what I was missing for the new hide feature working 100%. 

Now it works great, but there is an issue with the tests after adding the new migration for the hide column. See the failure in CI:
```
django.db.utils.ProgrammingError: column incident_incident.hide does not exist
```
Which also happens on my local when I run `python manage.py test`. I googled and started going down a path of resetting my migrations, and it got real weird on me, I don't know what to do with it. Let me know if you can help with this one. Thanks!

frontend/src/IncidentForm.js
```diff
@@ -37,6 +45,21 @@ const IncidentForm = ({ id, organization }) => {
...
   const mapRef = useRef(null);
 
+  const [showModal, setShowModal] = useState(false);
+
+  const handleClose = () => setShowModal(false);
+  const handleShow = () => setShowModal(true);
+  const handleConfirm = () => {
+    axios.patch(`/incident/api/incident/${id}/hide/`, { hide: true })
+      .then(() => {
+        window.location.href = `/${state.user.organization}`;
```
trevorskaggs, line 55: This line seems to push the user to `/undefined/`, but looks like the state object has drifted over time.  I think the problem might take a little work as we need to grab the organization `id` from the `state.organization` object, then do a look up for the index value of of the id in the `state.user.organizations` list, then use that index value to look up the `org_slug` from the `state.user.org_slugs` list for this redirect.

trevorskaggs (commented):
After fixing the redirect issue, we should be good to :shipit:

trevorskaggs:
Redirect Fixed!

PR #505, opened 2022-11-17
fixes team N increment #413

parksjr:
WIP:

This change fixes the Team N or Preplanned N default team name increment when planning a dispatch assignment. 

Does not introduce the new default team name function yet.

Closes #413

PR #720, opened 2024-04-18
Preplanned Team default name check + Team N increment #413

### LINKED ISSUE #413: When a DA is "preplanned" give the option to "Default" a new team name based on the next available number.
image
Add a "default name" button that will save the Team Name as the next Team X.

parksjr:
- Check that the "Team N" or "Preplanned N" Team name increment is working as expected.
  - always uses lowest available number.
  - had to change the endpoint that grabs the teams since it wasn't including preplanned teams for the name.
- The default name checkbox on the edit team name form dispatch summary page.
  - It disables the text box, and sends `{ defaultname: true }` as payload to the endpoint.
  - That property goes into the void as the backend isn't prepared for it.
  - I believe in the past (2 years ago!!), the plan was to have @AlexMountain implement something for the backend here. Let me know if we want to add it into the scope here, and if so need details on what it should do.

More notes

I closed the old pull request here #505 in favor of starting anew.

Closes #413

### DIFF
--- a/frontend/src/dispatch/DispatchMap.js
+++ b/frontend/src/dispatch/DispatchMap.js
@@ -248,30 +248,37 @@ function Deploy({ incident, organization }) {
           });
           setAssignedTeamMembers(response.data.filter(teammember => teammember.is_assigned === true).map(teammember => teammember.id))
           // Then fetch all recent Teams.
-          axios.get('/evac/api/dispatchteam/?incident=' + incident, {
+          axios.get('/evac/api/evacassignment/', {
             params: {
-              map: true
+              map: true,
+              incident,
             },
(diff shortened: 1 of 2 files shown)


trevorskaggs (commented):
Seems that the Deploy Teams functionality works, but I'm getting "Preplanned 1" for all the preplans still?

parksjr:
> Seems that the Deploy Teams functionality works, but I'm getting "Preplanned 1" for all the preplans still?

Good catch. The issue was after submitting a preplan, the number would go back to Preplanned 1. That is fixed, ready for your review again. @trevorskaggs

trevorskaggs (approved):
:shipit:

PR #750, opened 2024-06-06
folds whitespace in printed DA

### LINKED ISSUE #712: Compress Printed DA Spacing
When physically printing a DA, we end up with a lot of white space, let's compress the space to make for fewer total printed pages.

parksjr:
I've made a small change with a big impact regarding whitespace, there was a page break after the summary of each Service Request in a Dispatch Assignment, so I removed that, and tightened it up a little bit and at the same time making the font size of the headers a bit bigger to keep separation of sections. Screenshot comparison below (left: optimized; right: original). This is also very dynamic, but i feel this small change has the most sweeping impact.  If you want me to take it further into the  depths of the document, let me know and I will shrink it more.

[
Screenshot 2024-06-05 at 9 47 59 PM
](url)

Closes #712

### DIFF
--- a/frontend/src/dispatch/Utils.js
+++ b/frontend/src/dispatch/Utils.js
@@ -102,7 +102,8 @@ const buildDispatchResolutionsDoc = (drs = []) => {
       pdf.drawPad(10);
     });
 
-    // end of summary
+    // end of summary page break
+    pdf.drawPageBreak();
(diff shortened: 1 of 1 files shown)


parksjr:
@trevorskaggs I added the page break back in for this one 👍

PR #777, opened 2024-07-09
776: fixes animal link on MR to use `id_for_incident`

### LINKED ISSUE #776: Update MR Animal Link
Link back to Animal on MR page is to PK instead of incident id.

parksjr:
Fixes the animal link on the Medical Record (MR) Details to reference the `id_for_incident` property of the animal instead of the PK.

The medical record is fixed as requested by issue #776, however the same behavior exists on the Patient card component, which is present on the following pages:

- Diagnosis Form
- Diagnosis Results Form
- Exam Form
- Orders Form
- Procedure Results Form
- Treatment Plan Form
- Treatment Request Form

Should I apply the same fix to the Patient card component?

> Note: In my own testing I could not get the two numbers (`id` vs `id_for_incident`) to diverge, so technically I was unable to fully e2e test this, however, given the simple edit needed I figured that would be fine, and whoever reviews this would know how to test it proper. Let me know if I missed something.

Closes #776


trevorskaggs (approved):
🏰 🇲🇨

PR #847, opened 2024-09-18
Vet Treatment Plan Support

AlexMountain:


AlexMountain:
@trevorskaggs I'd like to keep adding to this but we can probably start reviewing the Treatment Plan functionality and other finished changes.

trevorskaggs:
Looking freaking awesome dude!  Small issues:
- When creating a new Medical Note, if you dont activate the date picker, it will pass initially a null and cause a BE error.
- While entering a Treatment administration, if you click "Now", then clear the field, you will get an error! "Bad Request"
- When editing a Treatment, if you make a change to a "future" Treatment (in my instance I was administering the next one that was due in an hour), then update the Treatment plan, those future treatments get wiped out! I think this is the right action for treatments that have not been touched.
- "Update Medical Plan" link doesnt work from MR page.
-

PR #858, opened 2024-10-21
Fix incident close for non-SuperUser

trevorskaggs:


trevorskaggs:

PR #879, opened 2024-11-11
Change how animal grouping is handled.

AlexMountain:


AlexMountain:

PR #881, opened 2024-11-13
Fix Login Issue

trevorskaggs:
Reverts trevorskaggs/shelterly#880


trevorskaggs (commented):
This seems to be working, will double check on UAT!

PR #930, opened 2025-01-14
Add animal wrong sr bug

trevorskaggs:


AlexMountain:
@trevorskaggs Let's review this one!

PR #981, opened 2025-02-12
Support Operation Service Requests that have no animals.

AlexMountain:


trevorskaggs:
Few issues I'm seeing:
1. When completing a DAR, the Operations SR should be able to be left in "Assigned", just not during "Save and Complete"
2. Trying to save a DAR with an Operations SR doesnt seem to work, even if SR is triaged to a possible option. No console error.
3. Trying to "save and complete" a DAR with an Operations SR doesnt seem to work, no console error.
4. During DAR, if UTC an OSR, get a form validation error:
image

AlexMountain:
@trevorskaggs Animal order fix will require you to run a migration.

trevorskaggs:
- Still not able to "save and close" an DAR with a OSR inside. Seeing the following console error:
```react-dom.development.js:88 Warning: React does not recognize the `showPassword` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `showpassword` instead. If you accidentally passed it from a parent component, remove it from the DOM element.```
- Somewhat convoluted path, but if you save an OSR as closed, then move to UTC (meaning the original entry was not valid), it is saving the the Close state when popping out of the DA.
- When removing the OSR from a current DA, it kicks the OSR into "cancelled" status
- In printed SR, should we put "N/A" in Animals section if there are no animals, to convey it is not bug?

PR #982, opened 2025-02-14
Fix visit note performance

trevorskaggs:


AlexMountain:

PR #1000, opened 2025-03-01
Hotline address rework

AlexMountain:


AlexMountain:
@trevorskaggs Should finally be ready for review/feedback!

AlexMountain:
@trevorskaggs Finally ready for some additional testing on this bad boy.

trevorskaggs:
Issues:
- When reusing an Existing SR, the Owner information is not populating from the current SR on the Owner Page
- If you change the Species of an existing animal, it throws an error.
- If you dont have multiple previous animals and dont click through "Next Animal" for each, it throws an error (might be the above issue true cause).  It also appears to potentially create a new SR and steal all the original animals.

Improvements:
- We might need to expose the Animal/Species counts in some capacity to help user select the right location if there are multiple at a location. Also helps to know what animals you might need to add if you get lost in the "add animal" workflow.

AlexMountain:
@trevorskaggs Ready for review with multiple owner support! Please also test the shelter intake workflow for bugs as I had to make some changes there to fix things.

PR #1026, opened 2025-03-14
Rework dispatch pages and deployment

AlexMountain:


AlexMountain:
@trevorskaggs Dispatch ticket is ready for review!

trevorskaggs:
This is looking good so far! Couple little things:
- When making a Preplan, it is using the Team name algo to give the next team name.
- Minor issue is the icon overlaps when filter is applied: 
image
- Did we have a regression were the DA Summary page is no longer sorting SRs? I thought I had the serializer kicking those back out by priority, then SR#, but it doesnt seem to be respecting that any longer.
image


Enhancement thoughts:
- Might be a lift, so we can push out, but showing the current pin locations (just little circles) when reusing a Team would be super-useful!
- Do we need to show how many items are currently filtered out?

AlexMountain:
@trevorskaggs Ready for another go. Only thing I couldn't reproduce/fix was the SR sorting on the DA summary PDF.

PR #1042, opened 2025-03-19
Fix DateRangePicker issues.

AlexMountain:


AlexMountain:

AlexMountain:
@trevorskaggs Good catch! Ended up making a migration to change disptach_date to just a DateField. Good for more testing!

PR #1097, opened 2025-05-01
Pagination for list views

AlexMountain:


AlexMountain:

trevorskaggs:
Sweet!
Couple notes:

Animal Search
- Species Advanced Filtering Throws an Error
- `ValueError: Field 'id' expected a number but got 'alpaca'.`
- Owned Advanced Filtering Throws an Error
- `django.core.exceptions.FieldError: Cannot resolve keyword 'owned' into field.`

Owners Search
- Doesnt seem to be showing all returned Owners, Reporters, or Owners/Reporters
- (Download says 9, but only showing 1)
image
- Combined:
image

Service Request
- Lost ability to hit enter on search bar
- When going directly to an SR#, it throws an error, if you clear the error the page loads:
Seems like this http request is finishing first and causing the issue: `/hotline/api/servicerequests/?light=true&incident=campfirecanbewa&organization=accorn`

Dispatch Assignment Search
- Filter by Open Date and Closed Date are date range widget instead of single date widgets!

Vet Search
- When selecting priority the Labels are showing var names:
image
- Procedures / Diagnostics: Ordered Date filter doesnt seem to work and cant apply with only a filter of Ordered Date
image
- Treatments: No Adminstrators in dropdown 
image

Holistically, can we take out the default advanced states for all searches? I think we have good dashboards/maps to represent those states now, and it is likely to be confusing if someone goes in and is "missing data". What do you think?

